PR #30, opened 2013-09-07
JERSEY-2085 fixed nested jar find class path resources.

otkmnb2783:
Nested jar files(e.g.jar://tmp/starter.jar!/libs/jersey-spring3.2.2.jar!/org/glassfish/jersey/server/spring /AutowiredInjectResolver.java) can find classmate resource.

https://java.net/jira/browse/JERSEY-2085


alexleigh:
I rebased this fix on top of latest Jersey master in #135.

PR #35, opened 2013-10-27
JERSEY-2171 jersey-spring: fix field injections on spring proxies

nikku:
- Fields get properly unwrapped before the spring component provider
  performs hk2 injection on them


trophies:
I ran into a similar issue as described in https://java.net/jira/browse/JERSEY-2171, I tested out this change and it resolves my issue (I filed a separate bug for the specific issue I had though - https://java.net/jira/browse/JERSEY-2919).

Is there any chance of this fix being included in a future jersey-spring3 release?  I attempted to write a test case for it, but the tests passed even without this fix in place.

bric3:
For those interested a pull request got merged that partly fix the problem : 
in #226 , relevant ticket https://java.net/jira/browse/JERSEY-3126

i.e. if you register an bean instance then injection won't happen, the patched code is not even called.

PR #76, opened 2014-04-08
Jersey 2469

EzequielB:
As UriBuilder spec defines, query parameters should be encoded using application/x-www-form-urlencoded rules (https://jax-rs-spec.java.net/nonav/2.0-SNAPSHOT/apidocs/javax/ws/rs/core/UriBuilder.html).

Changed UriComponent to conform to UriBuilder spec, making query parameter to be encoded correctly on three methods of UriBuilder (queryParam(...), replaceQueryParam(...), replaceQuery(...)).

Added new unit test to verify correct encoding of special characters, and adjust existent test to expect values encoded in query string. 

This set is expected to be encoded when used in query string: ! # $ & ' ( ) * + , / : ; = ? @ [ ];
 as defined in (http://en.wikipedia.org/wiki/Percent-encoding).

Check (https://java.net/jira/browse/JERSEY-2469 and https://java.net/projects/jersey/lists/users/archive/2014-04/message/13).


EzequielB:
Regarding the new info here (https://java.net/jira/browse/JERSEY-2469), 
Should we work on this and fix it @mpotociar?

mpotociar:
I have rebased this commit to the latest master as a pull request #183 and ran verification jobs internally. Closing this pull request as deprecated.

PR #88, opened 2014-06-20
getCookieStore method in ApacheConnectorProvider

eoff:
Following JERSEY-2424: after ApacheConnector became package-private we also lost access to CookieStore in ApacheConnector that is quite useful, for example, for QA issues etc. This small fix provides access to underlying CookieStore instance.


mgajdos:
Thank you for your contribution.

PR #90, opened 2014-07-03
Make the method really search for the @Path annotation in the ancestor h...

ccronemberger:
...ierarchy. Without this fix it is not possible to register proxied beans created by Spring because the annotation is in the super class

### DIFF
--- a/core-server/src/test/java/org/glassfish/jersey/server/model/internal/ModelHelperTest.java
+++ b/core-server/src/test/java/org/glassfish/jersey/server/model/internal/ModelHelperTest.java
@@ -37,49 +37,52 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-
 package org.glassfish.jersey.server.model.internal;
 
+import org.junit.Assert;
+import org.junit.Test;
+
 import javax.ws.rs.Path;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 
 /**
- * Common model helper methods.
- *
- * @author Michal Gajdos (michal.gajdos at oracle.com)
  * @author Constantino Cronemberger (ccronemberger at yahoo.com.br)
  */
-public final class ModelHelper {
+public class ModelHelperTest {
 
-    /**
-     * Get the class in the provided resource class ancestor hierarchy that
-     * is actually annotated with the {@link javax.ws.rs.Path &#64;Path} annotation.
-     *
-     * @param resourceClass resource class.
-     * @return resource class or it's ancestor that is annotated with the {@link javax.ws.rs.Path &#64;Path}
-     *         annotation.
-     */
-    public static Class<?> getAnnotatedResourceClass(Class<?> resourceClass) {
-
-        // traverse the class hierarchy to find the annotation
-        Class<?> cls = resourceClass;
-        do {
-            if (cls.isAnnotationPresent(Path.class)) {
-                return cls;
-            }
-
-            for (Class<?> i : cls.getInterfaces()) {
-                if (i.isAnnotationPresent(Path.class)) {
-                    return i;
-                }
-            }
-        } while ((cls = cls.getSuperclass()) != null);
-
-        return resourceClass;
+    @Test
+    public void testClass() {
+        Class cls = ModelHelper.getAnnotatedResourceClass(MyAnnotatedClass.class);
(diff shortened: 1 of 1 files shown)


bagana:
Hi there,

I found this fix by Googling about a different but probably related problem of jersey-spring3.

After hours of time debugging, I concluded that my `@Context`-annotated fields in Jersey resource and filter classes are ignored *once* I apply a Spring `@Transactional` on them. So I suspected that could it be Spring providing sub-classed instances to Jersey, and jersey didn't search the super classes for `@Context` annotation. I checked the diff of this fix but `@Context` seems to be handled somewhere else. I'm not familiar with Spring internal, nor with Jersey's. Can anyone give me some direction?

I tried with Jersey everything 2.19, Spring everything 4.1.6.

Thank you!

ccronemberger:
Hi,


That was exactly the use case I had. The workaround I used was to let Spring create the beans and register them on Jersey after the proxy is created. When I first implemented this I had to fix another problem in Jersey, but now the fix is already integrated on the latest release, so it should work the way I described above.


Let me know if you have further questions.


Regards,

    Constantino

Enviado do Yahoo Mail no Android

De:"Bagana" <notifications@github.com>
Data:11:47 dom, 12 de jul de AM
Assunto:Re: [jersey] Make the method really search for the @Path annotation in the ancestor h... (#90)

Hi there,

I found this fix by Googling about a different but probably related problem of jersey-spring3.

After hours of time debugging, I concluded that my @Context-annotated fields in Jersey resource and filter classes are ignored once I apply a Spring @Transactional on them. So I suspected that could it be Spring providing sub-classed instances to Jersey, and jersey didn't search the super classes for @Context annotation. I checked the diff of this fix but @Context seems to be handled somewhere else. I'm not familiar with Spring internal, nor with Jersey's. Can anyone give me some direction?

I tried with Jersey everything 2.19, Spring everything 4.1.6.

Thank you!

bagana:
Thanks for the reply and the idea, @ccronemberger.

I came up with this solution:

```java
	// "static", as long as it's singleton, should be fine.
	// @Component makes it singleton anyway.
	private static ResourceInfo resourceInfo;

	// "final", so it won't be overridden by Spring (cglib).
	// prints a warning message, though. (INFO level)
	@ Context
	public final void setResourceInfo( final ResourceInfo info) {
		resourceInfo = info;
	}

	@ Override
	public void filter( final ContainerRequestContext request) throws IOException {
		resourceInfo.getResourceMethod()...
		...
	}
```

So far it seems working fine. I guess this would probably be a better workaround, because if the problem should be solved in a future version, I only need to remove two key words.

mpotociar:
I'm afraid the solution in this patch is not correct. The JSR-339 specification in chapter 3.6 clearly states that annotations on a superclass take precedence over annotations on interfaces:

> JAX-RS annotations may be used on the methods and method parameters of a super-class or an implemented interface. Such annotations are inherited by a corresponding sub-class or implementation class method provided that the method and its parameters do not have any JAX-RS annotations of their own. Annotations on a super-class take precedence over those on an implemented interface. The precedence over conflicting annotations defined in multiple implemented interfaces is implementation specific. Note that inheritance of class or interface annotations is not supported.

ccronemberger:
The text that you mentioned says "Such annotations are inherited by a corresponding sub-class or implementation class method provided that the method and its parameters do not have any JAX-RS annotations of their own.".

The classes generated by Spring (not sure how) are sub-classes and they don't have any JAX-RS annotations (that is why I created this patch), so it should be OK. What this patch is doing is making the inheritance a reality.

ccronemberger:
Yes, I understand that inheritance is supported only for methods and method parameters and not for classes. The patch is for classes, so it is clearly violating the spec.
The option in this case would be to generate the subclasses with the annotations, so they don't need to be inherited.

AdamLindenthal:
Hi, Marek was originally pointing out, that the search order is opposite than the specs requires. The other thing is, that the spec states that "inheritance of class or interface annotations is not supported".

Well, if we should interpret that strictly, we would have to remove even the search in the interfaces already present before this patch. For now, we decided to interpret this statement more freely in the manner "this behaviour is neither specified, nor prohibited by the spec", so we just adjusted the patch so that it searches in the same order as specified for methods (superclasses first, interfaces if not found elsewhere).

Your use case should still work and Jersey behaves at least in the spirit of the spec...

Regards,
Adam

PR #94, opened 2014-07-21
Fix HttpDateFormat, so @BeanParam w/ @FormParam on Date type member can parse ISO dates

derekm:
Pull request for https://java.net/jira/browse/JERSEY-2542

My  name has been added to the OCA signatories list:
http://www.oracle.com/technetwork/community/oca-486395.html#m

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/HttpDateFormat.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/HttpDateFormat.java
@@ -109,7 +109,7 @@ public final class HttpDateFormat {
         @Override
         protected synchronized SimpleDateFormat initialValue() {
             final SimpleDateFormat format = new SimpleDateFormat(PREFERRED_DATE_FORMAT_PATTERN, Locale.US);
-            format.setTimeZone(TimeZone.getTimeZone("GMT"));
+            format.setTimeZone(TimeZone.getTimeZone("Etc/UTC"));
             return format;
         }
     };
@@ -130,7 +130,7 @@ public final class HttpDateFormat {
             new SimpleDateFormat(XSD_DATE_FORMAT_PATTERN_MIN2, Locale.US)
         };
 
-        final TimeZone tz = TimeZone.getTimeZone("GMT");
+        final TimeZone tz = TimeZone.getTimeZone("Etc/UTC");
         for (final SimpleDateFormat f : formats) {
             f.setTimeZone(tz);
         }
--- a/core-common/src/test/java/org/glassfish/jersey/message/internal/HttpDateFormatTest.java
+++ b/core-common/src/test/java/org/glassfish/jersey/message/internal/HttpDateFormatTest.java
@@ -69,7 +69,7 @@ public class HttpDateFormatTest {
     public void testPreferredFormatForParsingWithReferencePattern() throws ParseException
     {
         final SimpleDateFormat tester = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'");
-        tester.setTimeZone(TimeZone.getTimeZone("GMT"));
+        tester.setTimeZone(TimeZone.getTimeZone("Etc/UTC"));
         SimpleDateFormat preferred = HttpDateFormat.getPreferredDateFormat();
         date = new Date();
         final String today = tester.format(date);
@@ -89,7 +89,7 @@ public class HttpDateFormatTest {
     public void testPreferredFormatForFormattingWithReferencePattern()
     {
         final SimpleDateFormat tester = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'");
-        tester.setTimeZone(TimeZone.getTimeZone("GMT"));
+        tester.setTimeZone(TimeZone.getTimeZone("Etc/UTC"));
         final SimpleDateFormat preferred = HttpDateFormat.getPreferredDateFormat();
         date = new Date();
         assertEquals(tester.format(date), preferred.format(date));


AdamLindenthal:
Hi @derekm and @michael-o, 
honestly, I see this pull request being here for a very long time and I am sorry for that. 
Personally, I don't have any deep knowledge about this problematics, so before I (or someone else) will dive into this to verify, that we should merge this, it would really help us if you two would come to consensus on a final state. 
The last post of @michael-o looks like an open questions, could you please respond, @derekm?

Many thanks,
Adam

derekm:
Work has been very busy, that kept me from finalizing some last thing
agreed upon with @michael-o. I am about to start a new job, so should have
free time to finish this up. I am glad to hear continued interest. I will
make time today to research and commit the final determinations made
previously in this thread.

Thanks much, all!!

AdamLindenthal:
Hi Derek, 
many thanks, your effort is appreciated. 
And good luck on your new job.

Adam

derekm:
Okay, I've confirmed via tzdata zoneinfo source that "Etc/UTC" is the canonical name for UTC.

According to http://docs.oracle.com/javase/8/docs/api/java/util/Date.html, UTC and GMT are nearly equivalent, so the change to UTC should be acceptable. For the record, the canonical name for GMT is "Etc/GMT". In terms of zone definitions in the etcetera zoneinfo file, Etc/GMT, Etc/UTC and Etc/UCT appear to be equivalent in terms of offset and rules, although they are separately defined and not links to a unified parent zone.

I'm finishing the clean-ups according to @michael-o's findings and should have the commits posted soon.

derekm:
By the way, the day after @michael-o reported his findings, I set about confirming everything according to specs, and, as a result, I reported an errata to RFC 3339 related to decimal second values of less than 1. Just FYI.

Errata here: http://www.rfc-editor.org/errata_search.php?rfc=3339

I believe removing the incorrect patterns won't affect compliance with the corrected RFC.

derekm:
I had the incorrect rules to satisfy an example and a variant of that example from:
http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

The 3rd to last example under "Examples":
Pattern: "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
Result: 2001-07-04T12:08:56.235-0700

Being a not-to-spec, nonstandard example, it doesn't need to be satisfied.

I can switch the unit test to use the compliant 2nd-to-last example from the SimpleDateFormat documentation.

I also had the rules to satisfy PHP's invalid DateTime ISO 8601 pattern constant (http://php.net/manual/en/class.datetime.php):

    DateTime::ISO8601
    DATE_ISO8601
    ISO-8601 (example: 2005-08-15T15:52:01+0000)

Without the errant patterns, any PHP code that constructs form requests by using this constant will fail. Perhaps this single fact alone could provide support for an argument in keeping the errant patterns around, since this code is being enhanced primarily to allow unification of JSON/XML and x-www-form-urlencoded code paths (that is, JAXB @XmlElement annotation vs. BeanParam @FormParam annotation).

Otherwise, if we want to be to spec with regard to strict HTTP date headers parsing and lenient for x-www-form-urlencoded data while complying with invalid implementations like PHP, then code paths internal to Jersey may need some refactoring/splitting and the HttpDateFormat object maybe should not be used for both parsing needs.

I might vote for keeping the invalid patterns around to satisfy unification and popular bad actors like PHP.

Thoughts??

michael-o:
I am personally against supporting bad actors because this will also crawl into other systems/languages/frameworks, etc. In the case of PHP, it is plain wrong and needs to be fixed. If we say, we parse ISO 8601 it should be standards-compliant and reject any non-valid variant of it. Someone already [commented](http://php.net/manual/en/class.datetime.php#116085) on that, no reaction.

derekm:
I can agree with that! I'll get the remaining changes committed and pushed, and then this can finally be merged!

michael-o:
@derekm, great. I do not know how Jersey guys handle pull requests but we at Apache Maven prefer to have either one clean commit or logically compound commits. You change spreads over 10 commits, making it hard to read the log for one issue.

core-common/src/main/java/org/glassfish/jersey/message/internal/HttpDateFormat.java
```diff
@@ -129,15 +174,40 @@ public static SimpleDateFormat getPreferredDateFormat() {
...
+                latest = d;
+                last_pos = pp.getIndex();
+            }
+        }
+        if (latest != null) {
+            return latest;
         }
 
-        throw pe;
+        throw new ParseException("Unparseable date: \"" + date + "\"", -1);
```
mgajdos: Make sure to localize this message.

mgajdos:
I don't mind adding new date formats to list of dates Jersey is supposed to (un)marshall to/from. However the HttpDateFormat.readData is also used to read headers which is not the intended behavior you wanted to change. I'd suggest adding new method (with more date formats that would be used to process form params) and leave the old one to work with headers. The new method should be used in this param converter [1].

[1] https://github.com/jersey/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L246

AdamLindenthal:
Hi Derek, I can imagine you are busy, but do you thing you will have time to adjust the code according to Michal's comments? 
Thanks,
Adam

AdamLindenthal:
One more thing - before pushing new change, could you please rebase to the latest master? We had a change with relatively small impact, but on huge amount of classes, so this PR cannot be merged automatically in this state.

AdamLindenthal:
This pull request will soon have its first birthday :-)
Any updates, Derek? Everybody gave this up?

Please, let us know - we should either push it to the happy end or close it.

Thanks,
Adam

derekm:
I would very much like to get to @mgajdos recommendations. I've been in PHP-land with my new job, and that has unfortunately afforded little time to contribute to Java's state-of-the-art. I still think the problem this pull request solves is of great utility to Jersey, if only the technique it enables is popularized. I'll see what I can do this weekend, but if any paid contributor can get to it before I do, it wouldn't hurt my feelings in the least! :)

AdamLindenthal:
I see, thanks a lot. I don't think that anyone from the team will get to this in the near future, so your noone makes your contribution redundant in the meantime :-)

Cheers,
Adam

derekm:
I start a new job tomorrow back in the Java world. Here's hoping it gives me a chance to finish this pull request!

PR #99, opened 2014-08-18
Proposal to fix: #JERSEY-2490

athieriot:
Hi,

   I encountered a similar situation as explained here: https://java.net/jira/browse/JERSEY-2490?jql=project%20%3D%20JERSEY%20AND%20text%20~%20%22declarative%20link%22

   My case was an entity, managed by Hibernate, with a lazy loaded collection inside. 
   Even though I am not interested in loading this collection (Doing so by adding @XmlTransient on it), the DeclarativeLinked feature still try to recursively search for @Link fields in the said collection.

   Therefore, like explained in the ticket, I think it makes sense to have a way to bypass the recursive search of Jersey on demand.
   To do so, I propose a new annotation to mark a field we would like to ignore.

   The change is pretty simple actually.
   The name would probably need to be adapt a little bit more to the Jersey notation though.

Please tell me if I need to change some things (names...), write some documentation or anything else.
I am very keen to contribute the right way.

### DIFF
--- a/docs/src/main/docbook/declarative-linking.xml
+++ b/docs/src/main/docbook/declarative-linking.xml
@@ -262,6 +262,24 @@ List&lt;Link&gt; links</programlisting>
             Multiple link headers can be added by use of the <literal>@InjectLinks</literal> annotation
             which can contain multiple <literal>@InjectLink</literal> annotations.
         </para>
+      </section>
+
+    <section>
+        <title>Prevent Recursive Injection</title>
+        <para>
+          	By default, Jersey will try to recursively find all <literal>@InjectionLink</literal> annotations in
+          	the members of your object unless this member is annotated with <literal>@XmlTransient</literal>.
+
+          	But is some cases, you might want to control which member will be introspected regardless of the
+			<literal>@XmlTransient</literal> annotation.
+
+          	You can prevent Jersey to look into an object by adding <literal>@InjectLinkNoFollow</literal> to a field.
+
+			<programlisting language="java">
+				@InjectLinkNoFollow
+				Context context;
+			</programlisting>
+        </para>
     </section>
 
     <section>
--- a/docs/src/main/docbook/declarative-linking.xml
+++ b/docs/src/main/docbook/declarative-linking.xml
@@ -3,7 +3,7 @@
 
     DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
 
-    Copyright (c) 2010-2014 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2010-2015 Oracle and/or its affiliates. All rights reserved.
 
     The contents of this file are subject to the terms of either the GNU
     General Public License Version 2 only ("GPL") or the Common Development
@@ -267,11 +267,11 @@ List&lt;Link&gt; links</programlisting>
     <section>
         <title>Prevent Recursive Injection</title>
         <para>
-          	By default, Jersey will try to recursively find all <literal>@InjectionLink</literal> annotations in
+            By default, Jersey will try to recursively find all <literal>@InjectionLink</literal> annotations in
           	the members of your object unless this member is annotated with <literal>@XmlTransient</literal>.
 
-          	But is some cases, you might want to control which member will be introspected regardless of the
-			<literal>@XmlTransient</literal> annotation.
+          	But in some cases, you might want to control which member will be introspected regardless of the
+            <literal>@XmlTransient</literal> annotation.
 
           	You can prevent Jersey to look into an object by adding <literal>@InjectLinkNoFollow</literal> to a field.
 
--- a/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/FieldProcessor.java
+++ b/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/FieldProcessor.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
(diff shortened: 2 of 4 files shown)


AdamLindenthal:
Hi Atheriot, sorry for being inactive for some time regarding your pull request.
I wrote an email to kingsfleet and his point is clear now. 
With your test, you are proving, that the feature works when a field created by variable is annotated, such as:
```java
@InjectLinkNoFollow
private YourClass yourClass;
```

Gerard suggests there should be also test proving, that
```java
@InjectLinkNoFollow
private void setYourClass(YourClass instance) {
  // save the instance where ever you want
}
```

works as well. 
I did not read his original comment carefully enough. Now, after his explanation, I see that I should have understood it right away :-)

> Sorry I didn’t see the update. in the bean spec you don’t actually need a private field for a property you can just store it using a hashtable or similar. I was suggesting a test to make sure that the @NoLinkIntrospect will work on a field defined by accessors.
>
> Does this make more sense?
>
> Gerard

Would you be willing to add this?

Many thanks,
Adam

athieriot:
Hi Adam,

   Thank you for the following up. This is very much appreciated :)

   I am trying to get my head around this use case and, if I understand that you should be able to configure annotations or properties via accessors, I am not sure how this is suppose to work with the Link injection.

   In other words, before even talking about @InjectLinkNoFollow, I can't seem to reproduce a situation where Links are normally injected in a class affected via Setter or Getter.

   If I am doing something wrong, would you mind giving me an example please? I could probably add a section in the documentation then.

   If Jersey's declarative links feature does not support recursion over accessors, is it something you would like me to implement as part of this Pull Request?

Thank you !

Aurélien

docs/src/main/docbook/declarative-linking.xml
```diff
@@ -247,6 +247,24 @@ List&lt;Link&gt; links</programlisting>
...
         </para>
+      </section>
+
+    <section>
+        <title>Prevent Recursive Injection</title>
+        <para>
+          	By default, Jersey will try to recursively find all <literal>@InjectionLink</literal> annotations in
+          	the members of your object unless this member is annotated with <literal>@XmlTransient</literal>.
+
+          	But is some cases, you might want to control which member will be introspected regardless of the
```
shamoh: Type: `is` -> `in`.

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/InjectLinkNoFollow.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2010-2014 Oracle and/or its affiliates. All rights reserved.
```
shamoh: Please, change the year value just to 2015.

shamoh:
@kingsfleet, do you want to comment it more? I would agree with @athieriot that it is not subject of this PR to enhance Jersey Linking support to add accessors support, IMHO.

shamoh:
BTW, dear Universe, add to whitelist Mr. Aurélien Thieriot. Thanks.

docs/src/main/docbook/declarative-linking.xml
```diff
@@ -262,6 +262,24 @@ List&lt;Link&gt; links</programlisting>
             Multiple link headers can be added by use of the <literal>@InjectLinks</literal> annotation
             which can contain multiple <literal>@InjectLink</literal> annotations.
         </para>
+      </section>
```
shamoh: Update also year value of year in copyright header to ...-2015. Thanks.

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/FieldProcessor.java
```diff
@@ -50,6 +50,7 @@
 import java.util.Set;
```
shamoh: Update also year value of year in copyright header to ...-2015. Thanks.

incubator/declarative-linking/src/test/java/org/glassfish/jersey/linking/FieldProcessorTest.java
```diff
@@ -57,6 +57,7 @@
 import javax.ws.rs.core.MultivaluedMap;
```
shamoh: Update also year value of year in copyright header to ...-2015. Thanks.

docs/src/main/docbook/declarative-linking.xml
athieriot: Hello !

   I was going to update this only file but after checking, it appears that a lot (a lot !) of files still have the old copyright year all over the code base.

   Would you like me to update all of them in one commit or just the ones related to this PR?

Aurélien

shamoh: Just update year value of changed files.

kingsfleet:
Agreed this PR doesn't need to cover accessors. Sorry that was my confusion.

athieriot:
Hi,

   I've change the copyright dates on the files I modified to be 2015.

   I rebased my branch against master but I can't say what is happening with the tests. The test in error seems to be:
  
      ChunkedInputOutputITCase.chunkedOutputWithAsyncResponse:163 » Processing Timeo...

   Which has little to do with the Link feature.

Aurélien

shamoh:
@Jenkins, please retest this patch.
@kingsfleet Thanks.
@athieriot Thanks a lot. Don't worry about that, some tests are unstable. I will merge it manually it problem with build remains.

athieriot:
No worries.
Thanks to all of you for your time !

shamoh:
Jenkins, please test this patch .

shamoh:
Dear Jenkins, be so kind and test this patch again, please.

PR #102, opened 2014-09-04
Produce first declared content-type on Accept:*/*

mads-b:
This is a fix to reported issue JERSEY-2635, stemming from
the incorrect ordering of for loops in MethodSelectingRouter.


shamoh:
Hi Mads.
I think we can take your change without your commit history.
I would like to ask what was the reason you changed `methodBuilders` type from `Set` to `List`? Was there some special reason?
Thanks,
Libor

mads-b:
I'm not entirely sure, since it was a while ago I wrote this, but I recall fixing two bugs with this: This commit fixes the ordering of the values in @Produces annotation, but also an issue that surfaced after fixing this. For some reason, having multiple resource methods producing different media types results in random methods getting invoked when using the Accept: */* header, when the topmost method should be invoked. The reason it works without the patch is pure coincidence, but after fix, the problem became more prominent.

shamoh:
I've been 'inspired' by your code. Fix will be in 2.16 release, I hope. Thanks again for your contribution.

PR #111, opened 2014-09-25
Cache SSLSocketFactory so that connections can be reused. (JERSEY-2675)

rogern:
When using your own SSLContext in JerseyClient we found that the connections never were reused even though Keep-Alive was set. After some digging I found that the cache uses the SSLSocketFactory to lookup connections in the cache and that Jersey HttpUrlConnector always fetched the factory from the provided SLLContext and put it in HttpsURLConnection. 
The problem here was that the getSocketFactory method on SSLContextImpl always created new SSLSocketFactories and therefore connection lookup always failed. Hence new connections were created on each request.
This is a problem when calling resources far away (like across the Atlantic as in our case) where normal requests take approximately 200ms but with connection establishment and SSL handshake each request takes about 700ms.
This change will not enable this by any configuration, it will be the de facto setting. So if that would be a breaker I could make this configurable via a provider specific property and leave the old behaviour as default.

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnector.java
@@ -278,7 +278,7 @@ class HttpUrlConnector implements Connector {
             if (verifier != null) {
                 suc.setHostnameVerifier(verifier);
             }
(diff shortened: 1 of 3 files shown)


shamoh:
Universe, add to whitelist Mr. Roger. Thanks.

PR #114, opened 2014-10-09
[JERSEY-2680] Catch bubbling IllegalArgumentException

camann9:
Also catch IllegalArgumentException that can be thrown by UriTemplate which is called by JerseyUriBuilder. If there is an IllegalArgumentException during URL parsing it is very probable that the URL is malformed so I think catching is reasonable. This is also analogous to line 290 that has the same catch.

Jira-Ticket: https://java.net/jira/browse/JERSEY-2680

### DIFF
--- a/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/ServletContainer.java
+++ b/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/ServletContainer.java
@@ -330,7 +330,7 @@ public class ServletContainer extends HttpServlet implements Filter, Container {
             requestUri = absoluteUriBuilder.replacePath(requestURI).
                     replaceQuery(queryParameters).
                     build();
-        } catch (UriBuilderException ex) {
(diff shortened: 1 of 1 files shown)


mgajdos:
@camann9 In this [gist](https://gist.github.com/mgajdos/501c45d64a8539b825a1) you can find a test-case.

You covered only servlet part of the problem. If the Jersey application is deployed as servlet-filter then the change won't work. See [this](https://github.com/jersey/jersey/blob/master/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/ServletContainer.java#L530) line in ServletContainer.

Add the test to your change fix the servlet-filter issue and I'll merge the change. Thanks.

mgajdos:
Due to missing test-case the fix has been delivered via internal commit, see [1]. Thank you for noticing the problem.

[1] https://github.com/jersey/jersey/commit/47df9234de2fb9e6e24a30ded6ece10680c0c3c5

PR #131, opened 2014-12-07
JERSEY-2724 Added support for toString in proxy client

eiden:
If I call `toString()` on a proxy client, it throws an `UnsupportedOperationException`:

````
@Path("/")
interface MyClient {
    @GET String get();
}

@Test
public void buildClient() {
    WebTarget target = ClientBuilder.newClient()
                            .target("http://localhost/");

    MyClient client = WebResourceFactory.newResource(MyClient.class, target);

    client.toString(); //<-- throws java.lang.UnsupportedOperationException: Not a resource method
}
```
It's not a good thing to throw exceptions in `toString`. This PR add support for handling `toString` calls in the invocation handler.

Guava has an [AbstractInvocationHandler](https://github.com/google/guava/blob/fa95e381e665d8ee9639543b99ed38020c8de5ef/guava/src/com/google/common/reflect/AbstractInvocationHandler.java) where this logic is implemented (`hashCode` and `equals` as well). But the guava artifact is not in the proxy-client's pom. A better solution would be to use the guava's `AbstractInvocationHandler`, but I don't have the knowledge to decide on wether it's a good idea to add that dependency to the pom.

JIRA issue:
https://java.net/jira/browse/JERSEY-2724

### DIFF
--- a/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/WebResourceFactory.java
+++ b/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/WebResourceFactory.java
@@ -160,6 +160,10 @@ public final class WebResourceFactory implements InvocationHandler {
     @Override
     @SuppressWarnings("unchecked")
     public Object invoke(final Object proxy, final Method method, final Object[] args) throws Throwable {
+        if (args == null && method.getName().equals("toString")) {
+            return toString();
+        }
+
         // get the interface describing the resource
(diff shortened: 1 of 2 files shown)


eiden:
My name has made it's way into the OCA list. That only took a little under 2 months :)

eiden:
@AdamLindenthal I have rebased this to resolve a conflict

mgajdos:
@eiden 
Thank you for the contribution!

PR #137, opened 2015-01-23
[JERSEY-2762] Added support for @InjectLink and the @InjectLinks annotat...

carterlexy:
[JERSEY-2762] Added support for @InjectLink and the @InjectLinks annotations on methods

### DIFF
--- a/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
+++ b/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
@@ -41,6 +41,7 @@
 package org.glassfish.jersey.linking;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
@@ -52,7 +53,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
 import javassist.bytecode.SignatureAttribute.TypeParameter;
+
 import javax.ws.rs.core.Link;
 
 /**
@@ -84,6 +87,8 @@ class EntityDescriptor {
 
     private Map<String, FieldDescriptor> nonLinkFields;
     private Map<String, FieldDescriptor> linkFields;
+    private Map<String, MethodDescriptor> nonLinkMethods;
+    private Map<String, MethodDescriptor> linkMethods;
     private List<LinkHeaderDescriptor> linkHeaders;
 
     /**
@@ -102,6 +107,14 @@ class EntityDescriptor {
         findFields(entityClass);
         this.nonLinkFields = Collections.unmodifiableMap(this.nonLinkFields);
         this.linkFields = Collections.unmodifiableMap(this.linkFields);
+
+        // create a list of method names
+        this.nonLinkMethods = new HashMap<String, MethodDescriptor>();
+        this.linkMethods = new HashMap<String, MethodDescriptor>();
+        findMethods(entityClass);
+        this.nonLinkMethods = Collections.unmodifiableMap(this.nonLinkMethods);
+        this.linkMethods = Collections.unmodifiableMap(this.linkMethods);
+    
     }
 
     public Collection<FieldDescriptor> getLinkFields() {
@@ -115,6 +128,15 @@ class EntityDescriptor {
     public List<LinkHeaderDescriptor> getLinkHeaders() {
         return linkHeaders;
     }
+    
+    public Collection<MethodDescriptor> getLinkMethods() {
+        return linkMethods.values();
+    }
+
+    public Collection<MethodDescriptor> getNonLinkMethods() {
+        return nonLinkMethods.values();
+    }
+
 
     /**
      * Find and cache the fields of the supplied class and its superclasses and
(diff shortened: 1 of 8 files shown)


AdamLindenthal:

AdamLindenthal:
Hi Caterlexy,

Thanks,
Adam

carterlexy:
OCA has been approved and is listed on the Oracle OCA signatory page.

AdamLindenthal:
Hi Ryan, 
thank you for notifying. We will try to find some time to review and eventually merge your patch.

Many thanks so far,
Adam

AdamLindenthal:
Verified manually.

AdamLindenthal:
There have been some new checks added to the verification build, so I will try to trigger it once more.

AdamLindenthal:

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
```diff
@@ -190,4 +212,44 @@ private void findLinkHeaders(Class<?> entityClass) {
...
     }
+    
+    /**
+     * Find and cache the methods of the supplied class and its superclasses and
+     * interfaces.
+     * @param entityClass the class
+     */
+    private void findMethods(Class<?> entityClass) {
+        for (Method m: entityClass.getMethods()) {
+            InjectLink a = m.getAnnotation(InjectLink.class);
```
AdamLindenthal: Would it be possible to create a more general helper method, that would contain some common shared code? Basically from here to the end of loop, this is almost identical with `findFileds` (only the "output collection"is different), which is 100% understandable - it does the same thing as `findFields`, just for methods (except it doesn't need to look into superclasses and interfaces).
Just a suggestion, if it would make the code less readable, we can leave this way (after all, this is still in incubator).

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
```diff
@@ -190,4 +212,44 @@ private void findLinkHeaders(Class<?> entityClass) {
...
+                }
+            } else if (m.isAnnotationPresent(InjectLinks.class)) {
+                
+                if (List.class.isAssignableFrom(t)
+                        || t.isArray() && Link.class.isAssignableFrom(t.getComponentType())) {
+                    
+                    InjectLinks a2 = m.getAnnotation(InjectLinks.class);
+                    linkMethods.put(m.getName(), new InjectLinksMethodDescriptor(m, a2, t));
+                } else {
+                    throw new IllegalArgumentException("Can only inject links onto a List<Link> or Link[] object");
```
AdamLindenthal: If you decide to leave the method as it is, than it might make sense to "customize" the message to differentiate the message from the one with field injection.

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/InjectLinkMethodDescriptor.java
```diff
@@ -0,0 +1,139 @@
...
+     */
+    public InjectLink.Style getLinkStyle() {
+        return link.style();
+    }
+
+    /**
+     * TODO javadoc.
+     */
+    public String getLinkTemplate(ResourceMappingContext rmc) {
+        return InjectLinkFieldDescriptor.getLinkTemplate(rmc, link);
```
AdamLindenthal: I know, that this call is technically OK - it will work, but it does not really feel right. 

Actually the entire class again shares a huge amount of code with InjectLinkFieldDescriptor. I think it would really make sense to extract the common parts into abstract superclass or do some other refactoring that reduces the amount of repeating code.

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
carterlexy: I just spent some time looking at this to see if it was achievable and believe the code would be easier to maintain and more readable to leave as is. Each iterates over a separate collection as well and there is not a common superclass/interface for Method and Field that contains the information needed. For methods the return type is evaluated, for fields it's simply the type, for example. So a helper method would be filled with if statements and would become difficult to read/maintain I believe.

incubator/declarative-linking/src/test/java/org/glassfish/jersey/linking/MethodProcessorTest.java
```diff
@@ -0,0 +1,905 @@
...
+        }
+        
+        public void setRes2(URI uri) {
+        	this.res2 = uri;
+        }
+    }
+
+    @Test
+    public void testProcessLinks() {
+        System.out.println("Links");
```
AdamLindenthal: I know, that we have system.outs in some other tests as well, but I would prefer not to add new ones. We try to get rid of those, when we are occasionally changing some tests, so would be nice to use a jdk logger:
`private static final Logger LOG = Logger.getLogger(FooBarTest.class.getName())` and then simply `LOG.info()`, `LOG.fine()`, etc. 
Would be nice to have those "look, I am entering a method" log either removed completely or logged *bellow* the info level (even though I admit that you can definitely find some contra-examples in our existing code.

incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/EntityDescriptor.java
carterlexy: Done, message updated.

AdamLindenthal:
Hi, you are very fast to respond and created commit even before I scrolled at the end of the diff :-)

Tu summarise, the code seems to be functional and working. I would prefer to reduce the code repetition at least in the `InjectLinkMethodDescriptor`, also because the call to `getLinkTemplate()` in another class is a bit ugly. Other than that, there's nothing really serious and think it can be merged (it is in incubator after all). 
But I would definitely need you to rebase against latest master in order to be able to ever merge this.

carterlexy:
Thanks Adam, I'll look to eliminate some of the duplication that you pointed out and rebase. Shouldn't take too long, thanks!

AdamLindenthal:
Great, looking forward for that!

Thanks,
Adam

carterlexy:
Adam, this should be ready to go.

Thanks,
Ryan

AdamLindenthal:
Thanks, Ryan.
I will re-run the Jenkins verification build. It will most likely fail on `AsyncTest` which is not related to your patch, but we should at least see if all the other tests & checks passed fine.

AdamLindenthal:

AdamLindenthal:
And, after it passes, we'll need to do a quick follow-up review, but I hope that's quick this time.

AdamLindenthal:

PR #138, opened 2015-01-26
Make hasEntity() work with a closed stream.

dadowns:
InboundMessageContext.hasEntity() currently throws an IllegalStateException outside the try catch due to the extra call to entityContent.ensureNotClosed() if the entity stream is already closed.

This check is not needed here due to EntityInputStream.isEmpty() calling ensureNotClosed() internally.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/InboundMessageContext.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/InboundMessageContext.java
@@ -741,8 +741,6 @@ public abstract class InboundMessageContext {
      *         {@code false} otherwise.
      */
     public boolean hasEntity() {
-        entityContent.ensureNotClosed();
-
         try {
             return !entityContent.isEmpty();
         } catch (IllegalStateException ex) {
--- a/tests/e2e/src/test/java/org/glassfish/jersey/tests/e2e/client/ResponseCloseTest.java
+++ b/tests/e2e/src/test/java/org/glassfish/jersey/tests/e2e/client/ResponseCloseTest.java
@@ -49,6 +49,8 @@ import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.test.JerseyTest;
 
 import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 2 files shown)


AdamLindenthal:

AdamLindenthal:
Hi Dan,

Before we can review your pull request, we need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html). It usually takes some time until the OCA is officially approved and listed on the page linked above (which is a relevant list for us).

Also, every pull request should have a issue opened in our [bugtracker](https://java.net/jira/browse/JERSEY). It is a good practice to place mutually links from github to JIRA and vice versa.

Thanks,
Adam

dadowns:
In the process of getting the OCA submitted, will update when that's done.

dadowns:
Ok, OCA is in (Daniel Downs) and issue https://java.net/jira/browse/JERSEY-2843 has been created. Sorry for the delay on that.

AdamLindenthal:
Merged, thanks for your contribution and this time sorry for the delay on our side.

Regards,
Adam

dadowns:
Thanks!

mgajdos:
Hi @dadowns,

sorry to inform you but this change has to be reverted – it's against the JAX-RS specification. See Javadoc to [Response#hasEntity()](https://jax-rs-spec.java.net/nonav/2.0-rev-a/apidocs/javax/ws/rs/core/Response.html#hasEntity()).

Michal

dadowns:
If that's the case then why have a try catch that explicitly catches the IllegalStateException and returns false? If that's the spec then.... ok, but not having any "safe" way to check if there's an entity seems broken in that every call to hasEntity must be wrapped in the same try catch that's inside hasEntity.

Since the spec wants a IllegalStateException then the whole method can just be:

return !entityContent.isEmpty();

as internally it'll throw the exception when the connection is closed.

mgajdos:
Yeah, you're right about the simplification. Will look at it.

If you're not satisfied with what the spec offers, feel free to file an improvement request [here](https://java.net/jira/browse/JAX_RS_SPEC). Thank you!

PR #143, opened 2015-02-07
JERSEY-2776: Unquoting boundary string for multipart Content-Type headers

jontejj:
So that an apache-cxf client can be used from a jersey server.

The original problem is described by https://issues.apache.org/jira/browse/CXF-4504.

As that issue doesn't seem to be fixed anytime soon, this commit solves the issue in the mean time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1476,7 +1476,6 @@
                 <artifactId>pax-web-extender-war</artifactId>
                 <version>${pax.web.version}</version>
             </dependency>
-
             <dependency>
                 <groupId>org.openjdk.jmh</groupId>
                 <artifactId>jmh-core</artifactId>
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/internal/MultiPartReaderClientSide.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/internal/MultiPartReaderClientSide.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
@@ -62,7 +63,6 @@ import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.ContextResolver;
 import javax.ws.rs.ext.MessageBodyReader;
 import javax.ws.rs.ext.Providers;
-
 import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
@@ -75,7 +75,6 @@ import org.glassfish.jersey.media.multipart.MultiPart;
 import org.glassfish.jersey.media.multipart.MultiPartProperties;
 import org.glassfish.jersey.message.MessageBodyWorkers;
 import org.glassfish.jersey.message.internal.MediaTypes;
-
 import org.jvnet.mimepull.Header;
 import org.jvnet.mimepull.MIMEConfig;
 import org.jvnet.mimepull.MIMEMessage;
@@ -138,6 +137,7 @@ public class MultiPartReaderClientSide implements MessageBodyReader<MultiPart> {
         if (properties.getBufferThreshold() != MultiPartProperties.BUFFER_THRESHOLD_MEMORY_ONLY) {
             // Validate - this checks whether it's possible to create temp files in currently set temp directory.
             try {
+                //noinspection ResultOfMethodCallIgnored
                 File.createTempFile("MIME", null, tempDir != null ? new File(tempDir) : null).delete();
             } catch (final IOException ioe) {
                 LOGGER.log(Level.WARNING, LocalizationMessages.TEMP_FILE_CANNOT_BE_CREATED(properties.getBufferThreshold()), ioe);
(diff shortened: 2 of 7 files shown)


jontejj:
I'm signing a OCA, can I send it to any of the project leads here?

shamoh:
Thanks @jontejj, please send it to oracle-ca_us [at] oracle [dot] com. Thanks.

jontejj:
I've sent it now:)

mgajdos:
I don't quite understand what is your use-case. Jersey already supports case when a boundary parameter is quoted and boundaries separating body parts aren't (see [1]).

Your test sets up a Jersey application which is then approached by CXF client. The problem with this test is you change RuntimeDelegate to the one from CXF. This effectively changes behavior of Jersey server - the way Content-Type headers are read because they no longer are read by Jersey code but by CXF code. And CXF code doesn't remove quotes from header parameters like Jersey does. So in the end you had to modify media type in multi part reader because of the setup. But I don't understand why you need to have setup like this.

[1] https://github.com/jersey/jersey/blob/master/media/multipart/src/test/java/org/glassfish/jersey/media/multipart/internal/FormDataMultiPartReaderWriterTest.java#L622

jontejj:
Our use case is that we use the apache cxf client to consume http api:s from our jersey2 server. As we have apache cxf client first on the classpath, its runtime delegate is used. The multipart reader in jersey2 doesn't support this but the one in jersey1 does (http://grepcode.com/file/repo1.maven.org/maven2/com.sun.jersey.contribs/jersey-multipart/1.15/com/sun/jersey/multipart/impl/MultiPartReaderClientSide.java). With this fix jersey2 works even with the apache-cxf delegate, so essentially this is a regression bug. If jersey doesn't work when the wrong runtime delegate is set I think you should throw error at startup. As it's now it's completely silent and then random discrepancies like this happen.

jontejj:
I see that jersey2 handles:
        map.put(URI.class, _createHeaderDelegate(URI.class));
        map.put(Date.class, _createHeaderDelegate(Date.class));
        map.put(String.class, _createHeaderDelegate(String.class));

whilst cxf supports:
        headerProviders.put(Link.class, new LinkHeaderProvider());

the others are pretty much the same I'd think. Apparently it works for us without these as we don't use those header types. My main question is how things like this should be handled in the future. I find it kind of add that the apache cxf client has a runtime delegate that is only supposed to be used serverside. So perhaps I should report an issue there as well?

media/multipart/src/test/java/org/glassfish/jersey/media/multipart/internal/MultipartWithApacheCxfClientTest.java
```diff
@@ -0,0 +1,130 @@
...
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+
+package org.glassfish.jersey.media.multipart.internal;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.*;
```
mgajdos: No .* imports.

media/multipart/src/test/java/org/glassfish/jersey/media/multipart/internal/MultipartWithApacheCxfClientTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.glassfish.jersey.media.multipart.FormDataParam;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class MultipartWithApacheCxfClientTest extends MultiPartJerseyTest
+{
+	@Path("/inbox")
+	public static class JerseyMultipartResource
+	{
```
mgajdos: No { on a new line. And, please, use spaces instead of tabulators.

pom.xml
```diff
@@ -1489,7 +1489,13 @@
                 <artifactId>pax-web-extender-war</artifactId>
                 <version>${pax.web.version}</version>
             </dependency>
-
+            <!-- For testing multipart support with another http client -->
+	        <dependency>
```
mgajdos: Please, keep this dependency in the integration test only.

mgajdos:
OK, I see. Please, create an integration test (see [1]) instead of the one directly in the multipart module.

[1] https://github.com/jersey/jersey/tree/master/tests/integration

pom.xml
jontejj: Done.

jontejj:
See updated tests.

mgajdos:
Jenkins, test this patch.

mgajdos:
Please, apply this [1] diff to your change. It contains some copyright, pom, formatting fixes. I'll merge your change after the diff is applied and verified. Thank you.

[1] https://gist.github.com/mgajdos/290ee69a85447bc05af7

jontejj:
Applied diff.

mgajdos:
Jenkins, test this patch.

mgajdos:
Tried manual verification (should have done it before submitting the first patch) - it passed after applying the second patch, see [2]. Can you, please, apply that as well? This time it should be OK :)

[2] https://gist.github.com/mgajdos/290ee69a85447bc05af7#file-patch-2

mgajdos:
Verified manually.

mgajdos:
@jontejj 
Thank you for your contribution!

PR #145, opened 2015-02-21
Restore JarFileScanner support for parents with a trailing slash

Stephan202:
This is a follow up of #132. In that ticket @cguillot pointed out that as a side effect of the solution to #132, the `parent` parameter to `JarFileScanner` may no longer have a trailing slash. Since the `parent` parameter is clearly meant to be interpreted as a path component, I agree with @cguillot that this is a regression worth fixing.

This PR restores support for trailing slashes and adds two unit tests to prove it. As a side effect the `JarFileScanner` logic actually became quite a bit simpler.

As it stands, this PR introduces a small piece of conditional logic in the `JarFileScanner` constructor. If you prefer, I can also push this down into `#hasNext()` (at probably negligible performance cost).

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/scanning/JarFileScanner.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/scanning/JarFileScanner.java
@@ -75,7 +75,7 @@ public final class JarFileScanner implements ResourceFinder {
      */
     public JarFileScanner(final InputStream inputStream, final String parent, final boolean recursive) throws IOException {
         this.jarInputStream = new JarInputStream(inputStream);
-        this.parent = parent;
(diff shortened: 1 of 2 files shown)


mgajdos:
Jenkins, test this patch.

Stephan202:
Hmm, the Jenkins build failed due to a test timeout. Earlier builds failed for the same reason, so I assume this is unrelated. (Locally the whole build ran just fine.)

mgajdos:
Yeah, not caused by your change. This test sometimes fails on CloudBees. I'll verify this manually.

mgajdos:
Verified manually.

mgajdos:
Thanks for fixing the issue.

Stephan202:
No problem, you're welcome. Thanks for merging it :)

PR #146, opened 2015-02-26
Jersey 2625

kingsfleet:
A fix for issue JERSEY-2625, basically JPA was adding in a whole load of extra fields, the only way to really deal with this was to exclude all transient fields as they are not marked otherwise.

Whilst I was in there I also excluded a few other common cases such as String and primitive types in order to make things that much quicker.

### DIFF
--- a/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/FieldProcessor.java
+++ b/incubator/declarative-linking/src/main/java/org/glassfish/jersey/linking/FieldProcessor.java
@@ -40,12 +40,16 @@
 
 package org.glassfish.jersey.linking;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Modifier;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Set;
 import javax.ws.rs.core.Link;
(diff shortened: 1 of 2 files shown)


kingsfleet:
Jenkins, please test this patch

AdamLindenthal:

AdamLindenthal:
Hi Gerard, 
thank you for the patch.

Just to clarify - those Jenkins commands will only trigger the build if one of the defined group of Jersey developers writes it.

AdamLindenthal:
Hi, the build failure does not seem to be related to this PR, hence re-triggering.

AdamLindenthal:

AdamLindenthal:
Hi, even if it won't fix the ailing build (which does not seem to be related to your patch), could you please rebase to the latest master? There was a problem with jdk8 build fixed recently... 

Many thanks,
Adam

AdamLindenthal:
Verified manually, otherwise looks good to me. 
Well, maybe the message when the test fails could be a bit more conservative :-)

kingsfleet:
Adam, fair comment - I have updated the failure message sorry about that.
Jenkins, you are dead to me.

AdamLindenthal:
Gerard, many thanks. 
I hate to say that, but you've made a typo ("interator") when fixing that message... Normally I would merge it and create new internal commit to fix such a minor thing, but in this case - I would like to ask you to rebase to the latest jersey master, so you can fix that with the next commit. 

After this is done, I guess nothing prevents this PR from being merged.
Regards,
Adam

kingsfleet:
Done, I hope. Sorry about that.

AdamLindenthal:
Merged, thank you!

AdamLindenthal:
Gerard, please don't get confused by the OCA label, it servers just for our overview and unification.

Cheers,
Adam

PR #150, opened 2015-03-19
JERSEY-2818: Fixing RolesAllowedDynamicFeature to make it HTTP status code complaint, allowing both 401 and 403

brunosimioni:
Fixing RolesAllowedDynamicFeature to make it HTTP status code complaint, allowing both 401 and 403 HTTP status codes. 401 must return when Authorization header is absent or present with wrong pair of user/pass. 403 must return when server approves the Authorization header, but user is not allowed to access the current resource..

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/filter/RolesAllowedDynamicFeature.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/filter/RolesAllowedDynamicFeature.java
@@ -43,6 +43,7 @@ package org.glassfish.jersey.server.filter;
 import java.io.IOException;
 
 import javax.ws.rs.ForbiddenException;
+import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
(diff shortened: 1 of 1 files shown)


mgajdos:
Hi Bruno,

Before we can review your pull request, we need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html). It usually takes some time until the OCA is officially approved and listed on the page linked above (which is a relevant list for us).

Please, provide some tests to make sure the change is covered.

Thanks,
Michal

mgajdos:
Jenkins, test this patch.

mgajdos:
Delivered as an internal commit to make sure it'll be in 2.18. Closing.

adamretter:
This is broken in Jersey 2.22.2. The response is now always a HTTP 403, rather than either a 401 or 403. You can see that `throw new NotAuthorizedException("Should provide user/pass");` (https://github.com/brunosimioni/jersey/commit/2c0c27bc190f625f1af541acb148e3f8c86ebaa0#diff-872f4a95da4f26a5b4fa7873704e5bc7R136) has been replaced by `throw new ForbiddenException(LocalizationMessages.USER_NOT_AUTHORIZED());` (https://github.com/jersey/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/filter/RolesAllowedDynamicFeature.java#L139).

The fix was originally raised in https://java.net/jira/browse/JERSEY-2818 and then the change which undid the fix is in https://java.net/jira/browse/JERSEY-2908.

PR #166, opened 2015-05-15
Patch https clientserver grizzly

ophers:
This is a patch for [JERSEY-2836](https://java.net/jira/browse/JERSEY-2836)

### DIFF
--- a/examples/https-clientserver-grizzly/src/main/java/org/glassfish/jersey/examples/httpsclientservergrizzly/RootResource.java
+++ b/examples/https-clientserver-grizzly/src/main/java/org/glassfish/jersey/examples/httpsclientservergrizzly/RootResource.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2010-2012 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010-2015 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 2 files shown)


ophers:
Btw, I've just signed the Oracle Contributor Agreement and emailed it to oracle-ca_us [at] oracle [dot] com.

AdamLindenthal:
Hi Ophers, thanks for contributing. The OCA processing will take a while (but has to be done the first time only), so please be patient.

AdamLindenthal:

AdamLindenthal:
Hi again, I just triggered the verification build. Please don't panic if it fails (it most likely will). There is a problem, that does not relate to your patch.

AdamLindenthal:
Hi Ophers, I just realized, that you have not changed the copyright headers. If I merged your PR now, it would cause our future builds to fail until this is fixed.
Every time you touch a file, you need to check, that the copyright header contains the correct year.
In this case, please change `Copyright (c) 2010-2012` to `Copyright (c) 2010-2015`.

Thanks,
Adam

ophers:
OK.

AdamLindenthal:
Thanks, that looks good. Let's go through the (hopefully) last verification iteration. Overall I'm fine with the patch, so if it passes, I guess it can be merged.

AdamLindenthal:

AdamLindenthal:
Hi, the verification failed, but the problem is not related to your patch. It is a known problem, that should be already resolved. Hence, merging.

Thanks for your contribution!

PR #167, opened 2015-05-24
Fix for Versioned Applications - https://java.net/jira/browse/JERSEY-2626

lprimak:
New pull request fixes checkstyle errors

### DIFF
--- a/containers/glassfish/jersey-gf-ejb/src/main/java/org/glassfish/jersey/gf/ejb/internal/EjbComponentProvider.java
+++ b/containers/glassfish/jersey-gf-ejb/src/main/java/org/glassfish/jersey/gf/ejb/internal/EjbComponentProvider.java
@@ -40,6 +40,11 @@
 
 package org.glassfish.jersey.gf.ejb.internal;
 
+import com.sun.ejb.containers.BaseContainer;
+import com.sun.ejb.containers.EjbContainerUtil;
+import com.sun.ejb.containers.EjbContainerUtilImpl;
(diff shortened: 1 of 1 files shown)


lprimak:
OCA is approved for me already

AdamLindenthal:

AdamLindenthal:
Thanks Iprimak. We are a bit under load now, but we'll try to take a look at it soon.

lprimak:
Looks like it's ready to be integrated.
Thank you!

mgajdos:
Hi Lenny,

I know it's been a long time but would it be possible for you to create some integration tests that would verify it? An example of such a test can be seen here: https://github.com/jersey/jersey/tree/master/tests/integration/cdi-ejb-test-webapp

Otherwise it looks good.

Thanks,
Michal

lprimak:
I can do that but I would hope that it wouldn't hold up the patch integration. 
I.e. I can commit to doing the tests but please don't wait for them to integrate the patch. 

Current situation is a fatal error and the patch could not possibly be any worse since its been tested for over a year. 

Thank you.

mgajdos:
OK, we have a deal :-) Thank you for your contribution.

lprimak:
I looked at the test example provided, but unfortunately, it doesn't apply to this scenario.
The failure scenario is only relevant in deployment of versioned apps to GlassFish
$ asadmin deploy -name myapp:1.0
I don't see how I can simulate that in the Jersey integration tests.
And since all other tests pass, there are assurances no regression snuck in.

Not sure what else I can do.

PR #172, opened 2015-06-08
JERSEY-2880: Link HTTP Header: Headers may contain double-quote-less parameters.

sfuhrm:
Please see http://tools.ietf.org/html/rfc5988#section-5
which permits rel, rev, type and many more being without double quotes.

The jersey client doesn't accept valid server responses without this change.

See https://java.net/jira/browse/JERSEY-2880 for the according issue.


AdamLindenthal:
Hi Stephan, 

thanks again for the contribution. The same things apply as for your previous one - we will need you to sign the OCA and open the JIRA ticket (and, optimally, mutually link those, so that everything is easier to find).

Thanks,
Adam

AdamLindenthal:

mgajdos:
Thank you, @sfuhrm, for your contribution.

PR #175, opened 2015-06-23
JERSEY-2891: ExceptionMapper not chosen correctly in case of multiple ExtendedExceptionMappers with same generic type

kppk:
Current implementation ExceptionMapperFactory#find method used sorted map to store the mappers by distance of the generic exception type from the actual type. If there were two exception mappers with the same generic type, the mappers got overwritten in that map. Also I noticed that the exception mappers order was changing with server restart making the behavior unpredictable. 

This fix removed the sorted map, finds the mapper with the lowest distance doing the 'isMappable' check within the same loop.

### DIFF
--- a/core-common/src/test/java/org/glassfish/jersey/internal/ExceptionMapperFactoryTest.java
+++ b/core-common/src/test/java/org/glassfish/jersey/internal/ExceptionMapperFactoryTest.java
@@ -1,0 +1,550 @@
+package org.glassfish.jersey.internal;
+
+import org.glassfish.hk2.api.ActiveDescriptor;
+import org.glassfish.hk2.api.Descriptor;
+import org.glassfish.hk2.api.Filter;
+import org.glassfish.hk2.api.Injectee;
+import org.glassfish.hk2.api.MultiException;
+import org.glassfish.hk2.api.ServiceHandle;
+import org.glassfish.hk2.api.ServiceLocator;
+import org.glassfish.hk2.api.ServiceLocatorState;
+import org.glassfish.hk2.api.Unqualified;
+import org.glassfish.hk2.utilities.AbstractActiveDescriptor;
+import org.glassfish.jersey.spi.ExtendedExceptionMapper;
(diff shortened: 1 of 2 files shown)


mpotociar:

core-common/src/test/java/org/glassfish/jersey/internal/ExceptionMapperFactoryTest.java
```diff
@@ -0,0 +1,550 @@
+package org.glassfish.jersey.internal;
```
mpotociar: Missing copyright header. Please use a header e.g. from here: https://github.com/jersey/jersey/blob/master/core-common/src/test/java/org/glassfish/jersey/process/internal/ExecutorProvidersTest.java

core-common/src/test/java/org/glassfish/jersey/internal/ExceptionMapperFactoryTest.java
```diff
@@ -0,0 +1,550 @@
...
+            return Objects.hash(super.hashCode(), impl);
+        }
+    }
+
+    /**
+     * Stub implementation -- reason for this is to control the order of list of services in ServiceLocator.
+     *
+     * @param <T>
+     */
+    private static class ServiceHandleStub<T> implements ServiceHandle<T> {
```
mpotociar: I guess that better than mocking the locator, you can just configure one for testing purposes, e.g. like here:
https://github.com/jersey/jersey/blob/master/core-common/src/test/java/org/glassfish/jersey/process/internal/ExecutorProvidersTest.java#L238-L256

To simulate user-registered exception mappers (if needed), you can create bindings using the "@Custom" qualifier:
https://github.com/jersey/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/internal/inject/Custom.java

mgajdos:
Verified manually.

mgajdos:
Thanks for your contribution.

PR #178, opened 2015-07-08
JERSEY-2636: POST requests without Content-Type header pass through @…

gredler:
…Consumes check.

Three scenarios below have different behavior before/after this change:

@Consumes                         | Request Content-Type              | Old Result | New Result
----------------------------------+-----------------------------------+------------+-----------
text/plain                        | text/plain                        | 200        | 200
text/plain                        | image/jpeg                        | 415        | 415
text/plain                        | -                                 | 200        | 415
*/*                               | text/plain                        | 200        | 200
*/*                               | image/jpeg                        | 200        | 200
*/*                               | -                                 | 200        | 200
multipart/form-data               | multipart/form-data               | 200        | 200
multipart/form-data               | text/plain                        | 415        | 415
multipart/form-data               | image/jpeg                        | 415        | 415
multipart/form-data               | -                                 | 500        | 415
application/x-www-form-urlencoded | application/x-www-form-urlencoded | 200        | 200
application/x-www-form-urlencoded | text/plain                        | 415        | 415
application/x-www-form-urlencoded | image/jpeg                        | 415        | 415
application/x-www-form-urlencoded | -                                 | 500        | 415

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/MethodSelectingRouter.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/MethodSelectingRouter.java
@@ -234,7 +234,8 @@ final class MethodSelectingRouter implements Router {
          */
         boolean isConsumable(ContainerRequest requestContext) {
             MediaType contentType = requestContext.getMediaType();
-            return contentType == null || consumes.getMediaType().isCompatible(contentType);
+            return consumes.getMediaType().isCompatible(contentType)
+                    || (contentType == null && MediaType.WILDCARD_TYPE.equals(consumes.getMediaType()));
         }
 
         @Override
--- a/core-server/src/test/java/org/glassfish/jersey/server/internal/inject/FormParamTest.java
+++ b/core-server/src/test/java/org/glassfish/jersey/server/internal/inject/FormParamTest.java
@@ -440,6 +440,37 @@ public class FormParamTest extends AbstractTest {
         assertEquals("POST", responseContext.getEntity());
     }
 
+    @Path("/")
+    public static class ConsumesApplicationFormUrlEncoded {
+
+        @POST
+        @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+        public String post(@FormParam("foo") final String foo) {
+            assertEquals("bar", foo);
+            return "OK";
+        }
+
+    }
+
+    /**
+     * JERSEY-2636: POST requests without Content-Type header pass through @Consumes check
+     */
+    @Test
+    public void testConsumesMatching() throws Exception {
+        testConsumesMatching("application/x-www-form-urlencoded", 200);
+        testConsumesMatching("text/plain", 415);
+        testConsumesMatching(null, 415);
+    }
+
+    private void testConsumesMatching(final String contentType, final int expectedStatus) throws Exception {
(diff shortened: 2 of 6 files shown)


AdamLindenthal:

AdamLindenthal:
Hi Daniel, 

Before we can review your pull request, we need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html). Please note, that it takes some time until the agreement is officially approved and listed on the mentioned site (which is a relevant list for us).

Regards,
Adam

gredler:
OCA submitted last week, not sure when it'll show up on the website.

gredler:
My OCA was approved yesterday.

AdamLindenthal:
Hi, thanks for reminding us. 
I see, that the build has failed with two test failures - https://jersey-ci.ci.cloudbees.com/job/github-pull-request-verification/186/

Either the tests are wrong, or something needs to be changed with your patch. We will definitely take a look at it as well during review, but in the meantime, you can check this as well. Before merging we need all the tests to pass anyway.

Thanks,
Adam

gredler:
I've updated the two tests, but please double check the changes. Hopefully it's just a case of the tests relying on old/incorrect behavior, but in theory they could be indicative of issues with the proposed change.

stepanv:

AdamLindenthal:

stepanv:

stepanv:
Jenkins, please retest this patch.

AdamLindenthal:
Jenkins, please re-test this patch.

stepanv:
With the change you introduced, it seems the `org.glassfish.jersey.server.model.AsyncContentAndEntityTypeTest.testAsyncContentType()` test never ends. 
We cannot merge your pull request unless this problem is solved. I'll try to take a look at it but I don't promise anything...

gredler:
Thanks for the heads up! I had assumed that this was some pre-existing test fragility, but the test has now been adjusted to account for the new matching logic.

stepanv:

stepanv:
Jenkins, please re-test this patch.

stepanv:
Jenkins, you make me very very sad...

japod:

gredler:
Looking at the latest Jenkins build, it appears there are a ton of tests in the other modules that depend on the old matching behavior. Possibly an indication that the solution outlined in the [bug report](https://java.net/jira/browse/JERSEY-2636) is not the best approach?

stepanv:
Yeah, the problem is that we often have `GET` resource with `@Consumes` annotation in our tests in Jersey (`@Consumes` is usually declared at the class level). In these cases, we don't expect `415` if we don't send `Content-Type`. 
I'm still unsure what is the right interpretation of the specification.
I'm pretty sure, we should not make a difference between `GET` and `POST` or any other http methods when it comes to the matching algorithm.

gredler:
If you guys determine that the change to the matching algorithm is the right thing to do, I can go through and update all the affected tests in the other modules. But I'll hold off on that until there's a decision one way or the other :-)

stepanv:
So... I'm sorry, but we have to close this. Your work will be thrown away. 
I'll try to explain in https://java.net/jira/browse/JERSEY-2636 why we're staying with current implementation.
Anyway, thanks!

PR #179, opened 2015-07-09
JERSEY-2895: Log bad request reason.

lanxx019:
Due to recent upgrade from jersey1 to jersey2, we found out that Jersey actually eats the exception information when the request header is invalid. This normally is fine but the information from the `HeaderValueException` contains  valuable information for debugging purpose and should not be simply discarded. 

This PR simply log the error message with `FINE` level.

### DIFF
--- a/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java
+++ b/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java
@@ -409,6 +409,9 @@ public class WebComponent {
                 }
             });
         } catch (final HeaderValueException hve) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.log(Level.FINE, "Bad resuest: " + hve.getMessage());
+            }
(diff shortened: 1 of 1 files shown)


AdamLindenthal:
Hi Bin, 

thanks for creating the pull request.

Before we could review your pull request, we would need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) - either as an individual or on behalf of the company.

But.... I just did this quick fix on friday last week and it waits in our internal review tool, so I would suggest, that we do this internally, as the OCA process might take a month or so...

Regards,
Adam

lanxx019:
@AdamLindenthal, thank you. The internal route definitely works better. I do not want to wait a month for this. I will close this one.

AdamLindenthal:
Thanks, the change has been submitted and is now available in the snapshot version.
Regards,
Adam

AdamLindenthal:
Little correction - it is now available on *github*, not yet in the snapshot repo (this will take few hours, if everything runs smoothly).

Sorry for confusion.

lanxx019:
Cool, thank you very much for this.

PR #192, opened 2015-08-31
JERSEY-2405 - spring4 support

devsprint:
Added jersey-spring4 module. There is no code changes than spring3 module, just the dependencies are updated.


vetler:
Unnecessary code duplication, just exclude dependencies you don't want, and include your own.

devsprint:
replacing dependencies works just fine. I'm closing the PR as unnecessary.

PR #195, opened 2015-09-11
JERSEY-2728. Treat SSLException as connection failure

medb:


AdamLindenthal:

AdamLindenthal:
Hi Igor, 
thanks for your contribution. 

Before we can review your pull request, we need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html).

Also, for many good reasons, we have to require every patch to be backed by a test (one of the reasons being making sure, that we don't accidentally re-introduce the same bug as a side effect of another commit).

Regards,
Adam

medb:
Added tests and re-factored code to be more testable.
I have signed and send OCA on September 14th.

medb:

medb:
I just received notification that my OCA was processed

AdamLindenthal:
Hi, thanks for notifying. Seems like your name has been misspelled. Would you like to ping them to fix it?

dansiviter:
I can't help think this is making the code more brittle. I would say any `IOException` emanating from `workers#writeTo` or `setEntityStream` is a connection failure therefore, just catch that instead of the many verbose catch statements. Also, catching an `java.lang.Exception` is [bad form](https://community.oracle.com/docs/DOC-983543#catchingException) and should be avoided where possible.

pavelbucek:
I agree with @dansiviter .. I made some changes and merged this change to master branch (will push to 2.x as well).

Thanks for your contribution!

PR #202, opened 2015-11-12
Content-Length 0 case

hyeonjae:
The request with empty body, but Content-length set by zero

| Before | After |
|----------|---------|
| 411 error (Content length requried) | normal operation |

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
@@ -356,7 +356,9 @@ public class HttpUrlConnector implements Connector {
         secureConnection(request.getClient(), uc);
 
         final Object entity = request.getEntity();
-        if (entity != null) {
+        String contentLength = getContentLength(headers);
+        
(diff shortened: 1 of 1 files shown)


AdamLindenthal:
Hi 박현재,

Furthermore, nowadays we unfortunately cannot spend as much time with the PRs and review as we'd like to, so there are some delays to be expected even from our side.

Also to have the things nicely organised, every pull request should have a issue opened in our [bugtracker](https://java.net/jira/browse/JERSEY). And it is a good practice to place mutually links from github to JIRA and vice versa.

Thanks,
Adam

AdamLindenthal:
Hi hyeonjae, 

the pull request is sleeping here for quite a while already. 
Sorry for our delays - we've been a bit busy, but there's also no activity from your side.

Did you catch the message about the OCA? Are you willing to go through the process (it only needs to be done with the first pull request)?

If so, would you please mind adding a reproducer test case to support your patch? It's also to protect the fix to be reverted as a side-effect of another change (and to clearly justify the patch itself).

Many thanks,
Adam

hyeonjae:
I sent OCA to Oracle just before.

hyeonjae:
OCA Approved.

PR #203, opened 2015-12-11
Extended properties for MappedJettison wrapper

gary-harpaz:
Hi,

I would like to marshal JAXB JSON but without the root element wrapper as discussed in:
http://stackoverflow.com/questions/18417466/jaxb-remove-xmlrootelement-wrapper/18421298#18421298

I am using Jettison as my JAXB provider, unfortunately this is  not supported.
I have extended the JettisonConfig wrapper to include more properties to make this possible. 
I am now able to unmarshal without the root element wrapper using a context resolver such as:  

        public final class JaxbContextResolver implements ContextResolver<JAXBContext> {

            private final JAXBContext context;
            private final Set<Class<?>> types;
            private final Class<?>[] cTypes = {Flights.class, FlightType.class, AircraftType.class};

            public JaxbContextResolver() throws Exception {
            	System.out.println("Injected JaxbContextResolver");
                this.types = new HashSet<Class<?>>(Arrays.asList(cTypes));
                //this.context = new JettisonJaxbContext(JettisonConfig.DEFAULT, cTypes);
                this.context = new JettisonJaxbContext(JettisonConfig.mappedJettison().setDropRootElement(true).build(), cTypes);
            }

            @Override
            public JAXBContext getContext(Class<?> objectType) {
                return (types.contains(objectType)) ? context : null;
            }
        }

### DIFF
--- a/media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
+++ b/media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
@@ -42,9 +42,12 @@ package org.glassfish.jersey.jettison;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+
 
 /**
  * An immutable configuration of JSON notation and options. {@code JettisonConfig}
@@ -52,7 +55,7 @@ import java.util.Map;
  *
  * @author Jakub Podlesak (jakub.podlesak at oracle.com)
  */
-public class JettisonConfig {
+public abstract class JettisonConfig {
 
     /**
      * Enumeration of supported JSON notations.
@@ -75,17 +78,14 @@ public class JettisonConfig {
     }
 
     private final Notation notation;
-    private final Map<String, String> jsonXml2JsonNs;
-    private final List<String> serializeAsArray;
 
     /**
      * Builder class for constructing {@link JettisonConfig} options
      */
-    public static class Builder {
+    public abstract static class Builder {
 
         private final Notation notation;
-        protected Map<String, String> jsonXml2JsonNs = new HashMap<String, String>(0);
-        protected List<String> serializeAsArray = new LinkedList<String>();
+
 
         private Builder(final Notation notation) {
             this.notation = notation;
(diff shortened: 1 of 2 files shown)


gary-harpaz:
?

media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
```diff
@@ -112,8 +114,46 @@ private void copyAttributes(final JettisonConfig jc) {
...
+            super(Notation.MAPPED_JETTISON);
+            copyAttributes(defaultInstance);
+        }
+
+        private static final JettisonConfig.Mapped defaultInstance = new Mapped();
+
+
+
+        private void copyAttributes(final JettisonConfig.Mapped jc) {
+            xmlToJsonNamespaces = new HashMap<>(jc.getXml2JsonNs());
```
shamoh: Can you use same name for the attribute?

media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
```diff
@@ -112,8 +114,46 @@ private void copyAttributes(final JettisonConfig jc) {
...
+            serializeAsArray = new LinkedList<>(jc.getArrayElements());
```
shamoh: Can you use same name for the attribute?

media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
```diff
@@ -97,13 +97,15 @@ private Builder(final Notation notation) {
          * @return a non-null {@link JettisonConfig} instance
          */
         public JettisonConfig build() {
```
shamoh: What about making build an abstract method, remove strange `instanceof` a implement it in sub-classes?

shamoh:
Hi @gary-harpaz,

Furthermore, nowadays we unfortunately cannot spend as much time with the PRs and review as we'd like to, so there are some delays to be expected even from our side.

Also to have the things nicely organised, every pull request should have a issue opened in our [bugtracker](https://java.net/jira/browse/JERSEY). And it is a good practice to place mutually links from github to JIRA and vice versa.

And also provide unit tests.

Thanks,
Libor

shamoh:

media/json-jettison/src/main/java/org/glassfish/jersey/jettison/JettisonConfig.java
gary-harpaz: What do you mean?

shamoh: I mean:
```java
xmlToJsonNamespaces != xml2JsonNs
serializeAsArray != arrayElements
```

PR #207, opened 2015-12-27
JERSEY-2583: Reusable user token in ApacheConnector preventing SSL ha…

cfiehe:
…ndshakes on every request when client authentiation is required.


cfiehe:
I have corrected the ording issue, but don't see a way to add JUnits tests to the contribution that validate token reuse. The token (principal) is stored in the `HttpContext ` which is not accessible from the outside and I cannot find a way to add a request and response listener to the already built HttpClient.

cfiehe:
In order not to affect the normal behavior of the `ApacheConnector`, I have switched token reuse to `false` by default.

shamoh:
@cfiehe Please, provide test case.
@jerseyrobot, add to whitelist, test this patch, please.

cfiehe:
I have added two test cases which validate that enabling token reuse prevents multiple ssl handshakes across subsequent requests.

cfiehe:
I had to exclude `javax.servlet:servlet-api` from `org.apache.httpcomponents:httpclient` because of a version conflict with `commons-logging:commons-logging`. Jersey declares version `1.1`, but `org.apache.httpcomponents:httpclient` uses version `1.2`. Via `commons-logging:commons-logging:1.1` we get a transitive compile-time dependency to `javax.servlet:servlet-api:2.3` which causes `JettyHttpContainer` to fail during the test run.

Is it possible to upgrade `commons-logging:commons-logging` from version `1.1` to `1.2`.

Here is the current dependency tree without excluding `javax.servlet:servlet-api`:
```
[INFO] --- maven-dependency-plugin:2.10:tree (default-cli) @ jersey-apache-connector ---
[INFO] org.glassfish.jersey.connectors:jersey-apache-connector:jar:3.0-SNAPSHOT
[INFO] +- org.apache.httpcomponents:httpclient:jar:4.5:compile
[INFO] |  +- org.apache.httpcomponents:httpcore:jar:4.4.1:compile
[INFO] |  +- commons-logging:commons-logging:jar:1.1:compile
[INFO] |  |  +- log4j:log4j:jar:1.2.12:compile
[INFO] |  |  +- logkit:logkit:jar:1.0.1:compile
[INFO] |  |  +- avalon-framework:avalon-framework:jar:4.1.3:compile
[INFO] |  |  \- javax.servlet:servlet-api:jar:2.3:compile
[INFO] |  \- commons-codec:commons-codec:jar:1.9:compile
...
```

shamoh:
@cfiehe Yes, I think you could update `commons-logging` version.

cfiehe:
Jersey now uses `commons-logging:commons-logging:1.2`.

shamoh:
@jerseyrobot, test this patch.

cfiehe:
In have updated my commit in order to support the use case when different user tokens are used for different target hosts. I am using a `Cache` for storing the tokens, so that this feature does not cause a memory leak. A token expires at the latest after the last access plus the timeout value of the underlying `SSLSession`. This ensures that we do not remove user tokens for currently active sessions.

mikesteez:
This patch would be great! Any news on merging it to master?

PR #209, opened 2016-01-09
JERSEY-3023: Client proxy not marshalling object to json due to order…

psamsotha:
**JERSEY-3023: Client proxy not marshalling object to json due to order of arguments**

The original algorithm for creating the resource proxy loops through all the method parameters
and checks if the parameter contains an `@XxxParam` annotation. If it doesn't,
then it is determined to be the entity parameter. The problem with this is that
it doesn't take into account `@Context` and other possible annotation
that are not the entity, so the entity needs to always be the last parameter
in the event there are other parameters without `@XxxParam`.

There's not really much you can do about this, as the possibilities are limitless
as to what could cause a fault in this algorithm. A reasonable solution
(or maybe work around) this PR proposes is to just allow the developer to specify the entity
index explicitly in an overloaded factory method. Without using the overload,
this solution leaves the program to proceed as it normally would, so there is no 
effect to any previous usages.

Tests were added to verify the behavior of the entity index, in the test class
`WebResourceFactoryEntityIndexTest`.

Another thing I'd like to mention is that I added some documentation for the client proxy. I created it in a different branch, as I wasn't sure if this is something that should be part a different issue/PR. [This is the branch](https://github.com/psamsotha/jersey/tree/client-proxy-docs) that includes the documentation. What I added is appended to the end of the Client API documentation, as I felt this was the most appropriate location. I'd like to contribute this piece of documentation if it is satisfactory.

### DIFF
--- a/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/WebResourceFactory.java
+++ b/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/WebResourceFactory.java
@@ -96,7 +96,9 @@ public final class WebResourceFactory implements InvocationHandler {
     private final MultivaluedMap<String, Object> headers;
     private final List<Cookie> cookies;
     private final Form form;
+    private final int entityIndex;
 
+    private static final int NO_ENTITY_PARAM = -1;
     private static final MultivaluedMap<String, Object> EMPTY_HEADERS = new MultivaluedHashMap<>();
     private static final Form EMPTY_FORM = new Form();
     private static final List<Class> PARAM_ANNOTATION_CLASSES = Arrays.<Class>asList(PathParam.class, QueryParam.class,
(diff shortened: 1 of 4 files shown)


pavelbucek:
Hi Phnom,

thanks for your contribution. We cannot consider merging anything from you unless you sign OCA - Oracle Contributor Agreement, see http://www.oracle.com/technetwork/community/oca-486395.html.

Best regards,
Pavel

psamsotha:
Thanks @pavelbucek sorry for the late reply. It's been signed. I've had a couple email correspondence with a couple people at oracle and the last email I got said that the OCA was approved. I'm not sure if I'm supposed to provide you any information from those emails.

pavelbucek:
No, there is no other action required from you at this moment. I should be notified that you were added to the list (with your github username).

psamsotha:
@pavelbucek It was signed back when I made this PR, under the email paulsamsotha.dev at gmail.

PR #210, opened 2016-01-19
JERSEY-3038: Declarative Linking: Support @InjectLinkNoFollow on class level

weaselmetal:
This would be an okay fix I think


AdamLindenthal:
Hi weaselmetal,

thanks for contributing. Unfortunatelly, we cannot review & merge your PR right-off. 
Before we can, we need you to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html).

It takes some time before the OCA gets accepted, but it only needs to be done once (with the first pull request).

You can find some more details at: https://jersey.java.net/scm.html#/Submitting_Patches_and_Contribute_Code

Also, to protect your work from being reverted or broken as a side-effect of another change, we would like to ask you to back up the patch with some test.

Many thanks,
Cheers,
Adam

PR #211, opened 2016-02-09
JERSEY-3042: Resource scanning inefficient

lmprice:
This patch improves the performance of Resource merging when there are
large numbers of Resources with the same path. The runtime of the
mergeResources method has been greatly improved when there are > 100
matching paths, since the nested loop and element removals were refactored
out.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/model/Resource.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/model/Resource.java
@@ -48,6 +48,7 @@ import java.util.Set;
 
 import javax.ws.rs.Path;
 
+import jersey.repackaged.com.google.common.collect.ArrayListMultimap;
 import org.glassfish.jersey.Severity;
 import org.glassfish.jersey.internal.Errors;
 import org.glassfish.jersey.internal.util.collection.Value;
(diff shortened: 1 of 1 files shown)


lmprice:
This is my first contribution, and I wasn't quite sure how to proceed. Let me know if there's anything I'm lacking.

AdamLindenthal:
Hi Michael, thank you for contributing.

The process is a bit more demanding with the first contribution, we need you to sign the Oracle Contributor Agreement [OCA](http://www.oracle.com/technetwork/community/oca-486395.html). This has to be done the first time only. 

We also prefer, that each pull request has a JIRA issue opened, so that it is clear which problem the pull request tries to solve. As I can see, you've already done that, which is great. 

The last but not least thing - we (almost) strictly require, that each pull request is backed by at least one (unit or integration) test. This might look to be a bit harsh requirement, but it has deep practical impacts also for the contributors - this way we try to make sure, that we (or another contributor) don't accidentally waste their valuable work by changing the fixed behaviour back as a side-effect of another commit.

You can read more about contributing on our [website](https://jersey.java.net/scm.html).

And to be honest with you, I unfortunately have to add a little "disclaimer". Please be patient, the team is currently pretty busy, so it is very challenging to find some time to handle the process the pull requests. We are sorry for that, but there are delays from our side to be expected.

Thanks,
Adam

AdamLindenthal:

lmprice:
Hi Adam,

I actually submitted the OCA just before I submitted the patch. I'll look into adding a unittest for it. I'll have to think about that one. Since this is to fix a performance issue, it might be difficult to come up with a real test for it.

Thanks,

Michael

AdamLindenthal:
Hi Michael, I understand. 
It's probably not necessary to create a test, that would fail the build, if the performance is worse than expected. We rather have to make sure, that it does not break the functionality. 

So maybe try to thing if you can come up with something, that tests the very change you've done, if possible. If this is not reasonably doable, we won't insist on adding new test, that would not be reasonable.

The future reviewer will take a look how well do we have this area already covered by existing tests and if  it is enough to make the merging safe. If it is to complicated and not meaningful, feel free to explain everything you find important here in comments. Sorry for writing so generic, I didn't take a deeper look on your changes yet.

Regarding the OCA, it usually takes several weeks to get processed. We do usually not get notified, so feel free to ping us and we will change the label from "waiting for OCA" to "OCA".

Regards,
Adam

AdamLindenthal:

lmprice:
Looks like a check failed on the copyright header. It explicitly says not to alter it, so I'm not sure how to proceed.

AdamLindenthal:
Oh, I guess we should be more explicit about this in the contribution howto.
Yes, the copyright header should stay intact, but the copyright year should be always correct. 
New files should have only one number, now  `2016`, older ones a range, older ones a range, i.e. `2010-2016` in this case. In other words, the last number (regardless if there is one or two) has to match the year of last change.
Sorry for confusion.

Adam

PR #219, opened 2016-03-30
Make entity filtering work with maps

glucas:
Fixes JERSEY-3070.


pavelbucek:
Hello Gregory,

thanks for your contribution. Could you please include at least some sanity test with your change?

thanks,
Pavel

glucas:
Yes, I'll take a look at adding tests for the PRs I've submitted. Thanks.

glucas:
It looks like the only Jackson-specific tests that currently exist are in examples/json-jackson. I do see entity filtering tests in examples\entity-filtering, so perhaps I can add a Jackson-specific variant there to add test coverage for this change.

adrianboimvaser:
Is this dead?

glucas:
I haven't had a chance to add test coverage for this. The change itself is trivial but there didn't seem to be an obvious pattern for Jackson-specific tests in this area, and I have not yet had an opportunity to spend more time on it.

PR #221, opened 2016-04-14
Allow inheriting from FilteringJacksonJaxbJsonProvider

oillio:
I'd like to be able to inherit from this class so that I can add additional functionality to the provider, without having to duplicate the filtering code.

As background, I want to merge the Dropwizard custom provider functionality with this one.  As described in this comment:
https://github.com/dropwizard/dropwizard/issues/1341#issuecomment-210050810
Right now, I am inheriting from the Dropwizard provider and duplicating the code in FilteringJacksonJaxbJsonProvider.  I would prefer to reverse that, as the Dropwizard provider is considerably less complex.


AdamLindenthal:
Hi Dan,

thanks for contributing. 

I am sorry, but the proposed change doesn't sound like a good idea to me. We are speaking here about `internal.*` package which is not for "public" use, such as inheriting, etc - IOW it is nor API neither SPI.

I didn't check in detail what are you trying to do, but could it be achieved with a change in "public" classes (not `internal`), such as `org.glassfish.jersey.jackson.JacksonFeature`?

Anyway, to merge any pull-request, we need the first-time contributor to sign [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) (takes a while until tit is officially approved...) and we also prefer when every pull request has a issue opened in our [bugtracker](https://java.net/jira/browse/JERSEY). And we are also quite pedantic, that every patch is backed by at least some test...:-) But in this case, there's probably no added or fixed functionality, that could be tested.

Before going through the process of signing the OCA (if you are willing to do so), please try to come up with something cleaner than exposing internal package content.

Regards,
Adam

oillio:
Maybe this is a stupid question but, can we move `FilteringJacksonJaxbJsonProvider` to a non-internal package?
It doesn't appear to be using any, and it is not called by any, internal classes itself.

I don't think it would be feasible to change JacksonFeature or anything else in Jersey to provide this functionality.  The problem is that I need to use functionality from two `JacksonJaxbJasonProver`s.  This Filtering one and `JacksonMessageBodyProvider` from drop wizard.  As `JacksonJaxbJsonProvider` is designed to allow adding functionality through inheritance, my requested change would be the easiest solution.

It may be possible to modify Dropwizard to remove the need for registering its own `JacksonJaxbJsonProvider`.  I will look into solving this issue from that side.

PR #223, opened 2016-05-23
JERSEY-3008: Allow adding headers to OutboundEvent

brcolow:
This is a very simple, naïve attempt at fixing [JERSEY-3008](https://java.net/jira/browse/JERSEY-3008). The issue should explain the motivation behind wanting to be able to add headers directly to an `OutboundEvent` (for my specific needs I would use this to add a CORS header).

Any feedback would be greatly appreciated.

### DIFF
--- a/media/sse/src/main/java/org/glassfish/jersey/media/sse/OutboundEvent.java
+++ b/media/sse/src/main/java/org/glassfish/jersey/media/sse/OutboundEvent.java
@@ -43,6 +43,7 @@ import java.lang.reflect.Type;
 
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
 
 import org.glassfish.jersey.internal.util.ReflectionHelper;
 
@@ -61,6 +62,7 @@ public final class OutboundEvent {
     private final MediaType mediaType;
     private final Object data;
     private final long reconnectDelay;
+    private final MultivaluedMap<String, Object> eventHeaders;
 
     /**
      * Used for creating {@link OutboundEvent} instances.
@@ -74,6 +76,7 @@ public final class OutboundEvent {
         private GenericType type;
         private Object data;
         private MediaType mediaType = MediaType.TEXT_PLAIN_TYPE;
+        private MultivaluedMap<String, Object> eventHeaders;
 
         /**
          * Set event name.
@@ -241,6 +244,22 @@ public final class OutboundEvent {
         }
 
         /**
+         * Set any additional headers to be attached to the event.
+         *
+         * @param eventHeaders HTTP headers to add
+         * @return updated builder instance.
+         * @throws NullPointerException in case the {@code eventHeaders} parameter is {@code null}.
+         */
+        public Builder headers(MultivaluedMap<String, Object> eventHeaders) {
+            if (eventHeaders == null) {
+                throw new NullPointerException(LocalizationMessages.OUT_EVENT_DATA_NULL());
+            }
+
+            this.eventHeaders = eventHeaders;
+            return this;
+        }
+
+        /**
          * Build {@link OutboundEvent}.
          * <p>
          * There are two valid configurations:
@@ -261,7 +280,7 @@ public final class OutboundEvent {
                 throw new IllegalStateException(LocalizationMessages.OUT_EVENT_NOT_BUILDABLE());
             }
 
-            return new OutboundEvent(name, id, reconnectDelay, type, mediaType, data, comment);
+            return new OutboundEvent(name, id, reconnectDelay, type, mediaType, data, comment, eventHeaders);
         }
     }
 
@@ -282,7 +301,8 @@ public final class OutboundEvent {
                   final GenericType type,
                   final MediaType mediaType,
                   final Object data,
-                  final String comment) {
+                  final String comment,
+                  final MultivaluedMap<String, Object> eventHeaders) {
         this.name = name;
         this.comment = comment;
         this.id = id;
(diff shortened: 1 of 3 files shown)


AdamLindenthal:
Hi Michael,

many thanks for your contribution and sorry for the delay, we have been busy on other stuff recently.

Thanks,
Adam

AdamLindenthal:
Hi again, one more thing... Would you be able to come up with a unit test supporting your solution? We usually insist on it - it serves also to protect your precious work from being removed/broken as a side-effect of another change.

I know, that the process is quite complex especially for the first time (OCA), sorry for that...

Regards,
Adam

brcolow:
@AdamLindenthal Thanks for the feedback, I will get it done.

AdamLindenthal:
Great, thanks.

brcolow:
I e-mailed the signed OCA today and also updated the PR to add a test.

delabassee:
OCA reviewed and approved

brcolow:
CA is signed, can we move forward?

brcolow:
Ping @pavelbucek

brcolow:
Could I get this reviewed?

pavelbucek:
Do I understand correctly that you are trying to add HTTP header to individual event (message) in SSE stream? If so, then that's not correct, since the headers are sent only once and there can be "n" events in single stream.

brcolow:
Yes..the intent is to add headers to each individual event message. Any idea where that is in the code base?

pavelbucek:
that is simply not possible, see SSE protocol definition: https://html.spec.whatwg.org/multipage/comms.html#server-sent-events

Individual events can have "id", "data", "event" and "retry". That's it. HTTP response headers are sent only once, with initial response.

brcolow:
Okay, thanks for the information. Would it be possible for you to read the original issue and ascertain if the original intent would be satisfied by sending the headers only once with the initial response, or if it is simply incompatible with SSE. I am sorry for taking up your time on this matter, but your expertise is highly appreciated.

pavelbucek:
Well, I don't see why you'd need to change the code. You should be able to achieve the same thing with a ContainerResponseFilter (just check for the content type, and if it's text/event-stream, add the header).

Or if you want to set the header just for the single stream (not event!), you can return a response (instead of EventOutput), set the header and set EventOutput as an entity, something like:

```    @GET
    @Path("events")
    @Produces(SseFeature.SERVER_SENT_EVENTS)
    public Response itemEvents(
            @HeaderParam(SseFeature.LAST_EVENT_ID_HEADER) @DefaultValue("-1") int lastEventId) {
        final EventOutput eventOutput = new EventOutput();

        // ...
        
        return Response.ok(eventOutput, SseFeature.SERVER_SENT_EVENTS_TYPE)
                       .header("myHeader", "myValue")
                       .build();
    }
```

(untested code)

brcolow:
I am trying this with the following code:

```java
@Path("twofactor")
@Singleton
public class AccountLinkService
{
    private SseBroadcaster broadcaster = new SseBroadcaster();
    
    @POST
    @Path("/notify-linked")
    @Produces(MediaType.APPLICATION_JSON)
    @Consumes(MediaType.APPLICATION_JSON)
    public Response broadcastAccountLinked(ObjectNode accountLinkedJson)
    {
        OutboundEvent event = new OutboundEvent.Builder().name("account-linked")
                .mediaType(MediaType.APPLICATION_JSON_TYPE)
                .data(ObjectNode.class, new ObjectNode(JsonNodeFactory.instance).put("userId", 2)).build();

        broadcaster.broadcast(event);
        
        return Response.ok(new ObjectNode(JsonNodeFactory.instance).put("success", true)).build();
    }

    @GET
    @Path("/link-events")
    @Produces(SseFeature.SERVER_SENT_EVENTS)
    public Response listenToAccountLinkedBroadcasts(@HeaderParam(SseFeature.LAST_EVENT_ID_HEADER) @DefaultValue("-1") int lastEventId)
    {
        final EventOutput eventOutput = new EventOutput();
        broadcaster.add(eventOutput);
        return Response.ok(eventOutput, SseFeature.SERVER_SENT_EVENTS_TYPE)
                .header("Access-Control-Allow-Origin", "*")
                .header("Access-Control-Allow-Headers", "origin, content-type, accept, authorization")
                .header("Access-Control-Allow-Credentials", "true")
                .header("Access-Control-Allow-Methods", "GET, POST")
                .build();
    }    
}

@Provider
public class CORSResponseFilter implements ContainerResponseFilter
{

    public void filter(ContainerRequestContext request, ContainerResponseContext response) throws IOException
    {
        if (response.getMediaType().equals(new MediaType("text", "event-stream")))
        {
            response.getHeaders().add("Access-Control-Allow-Origin", "*");
            response.getHeaders().add("Access-Control-Allow-Headers", "origin, content-type, accept, authorization");
            response.getHeaders().add("Access-Control-Allow-Credentials", "true");
            response.getHeaders().add("Access-Control-Allow-Methods", "GET, POST");
        }
    }
}
```

However, attempting to consume the event source using Javascript leads to an empty response error, and indeed attempting to `curl localhost:4463/twofactor/link-events` results in:

```
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 4463 (#0)
> GET /twofactor/link-events HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:4463
> Accept: */*
>
* Empty reply from server
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server
```

I can't tell if the CORS headers work as the method seems to be busted...

brcolow:
That works perfectly for my use case. Thanks *so* much.

pavelbucek:
perfect. Closing the pull request, since I don't think this should be merged.

PR #226, opened 2016-06-22
Fix Spring AOP and @Context not being injected

max-norris:
Fixes the issue where if you have @Context and an aspect on a resource, the fields do not get injected, see: https://java.net/jira/browse/JERSEY-3126.

Example project with a typical application: https://github.com/max-norris/jersey-fix-example.  It only works when you build the snapshot locally and then change the version in the POM.


pavelbucek:
I squashed the commits to the single one, made some cosmetic changes and merged to 2.x and master branches (2.x commit: https://github.com/jersey/jersey/commit/33d07ca7ef9e198fba304b73c91bc30e3b435f3c).

thanks for your contribution!

bric3:
This partly fix the problem ; if the code register a bean instance then Context injection won't happen, the patched code is not even invoked.

PR #227, opened 2016-07-04
[JERSEY-3134] Improve OAuth2-client support by providing design for additional flows. Add client credentials grant workflow

deepakpol:
* Changed design for more flexible OAuth2 grant flow implementations
* Added implementation for authorization code, and client credentials flows based on new design
* Added/modified e2e tests accordingly

Compatibility note: The feature is not fully compatible with old implementation for oauth2-client. The changes being return types for `OAuth2ClientSupport.authorizationCodeGrantFlowBuilder` which returns more generic `OAuth2InteractiveWorkflow.Builder` rather than specific to auth code workflow. Also the method names to execute the workflow have been changed to `execute()` and `resume()` instead of `start()` and `finish()`


pavelbucek:
Hi Deepak,

thanks for your contribution. We cannot consider merging anything from you unless you sign OCA - Oracle Contributor Agreement, see http://www.oracle.com/technetwork/community/oca-486395.html.

And about the pull request itself - seems like you are changing Jersey public API, which should be done in backwards compatible manner (or we can deprecate, release, remove, but it cannot be done in a single change).

Thanks and regards,
Pavel

deepakpol:
@pavelbucek thanks for your response. Before signing OCA, I would like to ensure that the changes are inline with what Jersey community wants and acceptable. From that regard, are you suggesting that the old classes and APIs be left there for now in `@Deprecated` state?

manhhavu:
@deepakpol : it has been a year since your pull request was made. It seems that you have some question before the OCA? Could you maybe contact @mpotociar (as he is currently assigned in the issue  #3405)?.

As I wish to have this feature, I'd like to know whether the pull request will be rolled before rolling my own one. 

Thank you,

Manh-Ha

PR #231, opened 2016-08-05
2.x

tantk:


pavelbucek:
Please provide some description of intended change.

Closing for now, since it looks like a user error.

PR #237, opened 2016-09-14
Refactored selection of exception mapper to take priority into account

israelchernyak:
Changing the selection of the exception mapper so that priorities (from the @Priority annotation) are taken into account

This is related to the following Jira: https://java.net/jira/browse/JERSEY-2437


blommish:
Does this have any tests?

pavelbucek:
already implemented.

PR #241, opened 2016-10-03
Support custom cookiespec registries and cookie stores in ApacheConnector

onmishkin:
Expose the underlying Apache HttpClient support for custom cookiespec registries and cookie stores to users of the Jersey API. Modeled after similar existing exposures (e.g., of HttpClient's REQUEST_CONFIG property).


connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -198,6 +202,7 @@
     private final CookieStore cookieStore;
     private final boolean preemptiveBasicAuth;
     private final RequestConfig requestConfig;
+    private final Lookup<CookieSpecProvider> cookiespecRegistry;
```
jgdef: I don't know if "cookieSpecRegistry" is the correct came case form? Given the placement of underscores in COOKIE_SPEC_REGISTRY.

jgdef:
:shipit:

PR #188, opened 2015-08-17
Compilation fixes for JERSEY-2940: Support Jetty 9.3

azell:
Code changes for https://java.net/jira/browse/JERSEY-2940

PR #243, opened 2016-10-05
JERSEY-2940 Allow runtime compatibility with jetty 9.3

jbarnettwomply:
This change still compiles against jetty 9.2, but allows runtime
compatibility with jetty 9.3 in the case we're hitting

### DIFF
--- a/containers/jetty-http/src/main/java/org/glassfish/jersey/jetty/JettyHttpContainer.java
+++ b/containers/jetty-http/src/main/java/org/glassfish/jersey/jetty/JettyHttpContainer.java
@@ -163,7 +163,7 @@ public final class JettyHttpContainer extends AbstractHandler implements Contain
     public void handle(final String target, final Request request, final HttpServletRequest httpServletRequest,
                        final HttpServletResponse httpServletResponse) throws IOException, ServletException {
 
-        final Response response = Response.getResponse(httpServletResponse);
+        final Response response = request.getResponse();
         final ResponseWriter responseWriter = new ResponseWriter(request, response, configSetStatusOverSendError);
         final URI baseUri = getBaseUri(request);
         final URI requestUri = getRequestUri(request, baseUri);


mjgp2:
Would be great to get this merged in.

@jbarnettwomply did you just end up building your own jar and putting on a maven repo?

jbarnettwomply:
yes, we built our own jar and put it in our internal maven repo.

defnull:
Jetty 9.4 has the same problem. A merge would be great.

tatacalu:
Hey guys, any update on the decision to merge this pull request?

pavelbucek:
Thanks for your contribution.

Also, wouldn't updating the Jetty dependency better option? (it would most likely need some change in the implementation, I assume this PR is the consequence of some backwardly incompatible API change in Jetty codebase).

Best regards,
Pavel

jbarnettwomply:
Updating the Jetty dependency was ruled out in the tagged jira before I looked at this issue due to its java 8 hard dependency.

Will work on getting OCA sent in.

jbarnettwomply:
@pavelbucek OCA has been approved.

pavelbucek:
Jetty version needs to be updated anyway because of PATCH method support in the connector. Java SE 8 is no longer an issue.

Closing the PR, expect this to be done in the master branch when new JAX-RS API milestone is uptaken.

jbarnettwomply:
+1 on updating jetty.  in the meantime, would it be possible to put this in a 2.25.x release?  We uptake jersey through dropwizard, and it would be great to not have to mess with our own versions of dependencies while a release from master goes through.

pavelbucek:
2.25.x is reserved for bugfixes in upper stack products and other priority escalations, which means that we don't really want to push any other commits there. So for now, 2.26 is the next stable Jersey to be released.

jbarnettwomply:
got it.  could this (or something like https://github.com/jersey/jersey/pull/188 if you jdk8 is no longer an issue) make it into 2.26 then?

pavelbucek:
it is already there ;-)

(as I mentioned above - it was done as part of migrating to latest JAX-RS 2.1 milestone: https://github.com/jersey/jersey/commit/b62aa3fb8f787fba54868e5346a837d661c2f500)

PR #254, opened 2017-01-19
Expose copy constructor of JerseyUriBuilder

tareqhs:
Allow extensions to extend ``JerseyUriBuilder`` and make use of this efficient copy constructor instead of wastefully re-creating a new one from scratch ( ``new`` then ``uri(..)`` ). Forcing to re-create``JerseyUriBuilder`` extensions using ``uri(..)`` causes a critical performance bottleneck in my scenarios.


pavelbucek:
Hi Tareq,

thanks for your contribution!

To be able to process your pull request, you have to sign OCA: http://www.oracle.com/technetwork/community/oca-486395.html.

Also, you are mentioning performance impact in "your" scenario. Can you please share the usecase and your measurements?

Thanks and regards,
Pavel

PR #255, opened 2017-01-20
Add callback hook for modifying the invocation.

daicoden:
This is intented to be used when dealing with APIs which use the entity
body to pass metadata.  As an example, json-rpc (http://json-rpc.org/)
specifies a protocol where the url remains constant, and a paramater
"method" determines the action performed.  That means you can't leverage
types because you end up an API that looks like this...

public class FreeIpaClient {
  @Consumes("application/json")
  @Produces("application/json")
  @POST @Path("/ipa/sessions/json")
  Response doRpc(JsonRpcCall call);
}

All leverage is lost.  In order to build a tool which is able to execute
different json-rpc actions based on the interface, there needs to be
some scope where the invocation context and the invoked method are
available.  This change creates that scope.  You can now have the
following.

public class FreeIpaClient {
  @Consumes("application/json")
  @Produces("application/json")
  @POST @Path("/ipa/sessions/json")
  @CustomJsonRpcAnno("user_show")
  UserShowResponse userShow(UserShowParams params);

  @Consumes("application/json")
  @Produces("application/json")
  @POST @Path("/ipa/sessions/json")
  @CustomJsonRpcAnno("user_delete")
  UserDeleteResponse userDelete(UserDeleteParams params);
}

With this interface, you can set a property on the Invocation.Builder
which contains the annotation information from the invoked method
userShow or userDelete.

You still need to do work to register a message body write/reader to
append the metadata contained in the annotation.  If desired you could
even do reflection based off of the method name.

I had a rough time with the tests - happy to re-work them.


daicoden:
Also happy to accomplish this in a different way if there are suggestions.

pavelbucek:
I'm slightly confused - you are introducing WebResourceInvocationCallback, but you are never implementing it.. can you please explain in more detail?

Also, for us to be able to process any pull request, we need author to sign the OCA: http://www.oracle.com/technetwork/community/oca-486395.html.

Thanks and regards,
Pavel

daicoden:
Hey @pavelbucek, signed the OCA.

Regarding the WebResourceInvocationCallback.  It's only implemented in tests to demonstrate how it could be used. It provides an extension API to allow further customization of a request off ofthe client interface.  It provides a scope in which both the invoking method is available for reflection *and* the request context.

Another solution would be to set a property on the builder with the invocation information, like `org.glassfish.jersey.client.proxy.InvocationData`.

Other suggestions welcome too!

Concretely, The problem that I have is - how do you attach metadata, known at compile time, to the request body?

In my example:
```
public class FreeIpaClient {
  @Consumes("application/json")
  @Produces("application/json")
  @POST @Path("/ipa/sessions/json")
  @CustomJsonRpcAnno("user_show")
  UserShowResponse userShow(UserShowParams params);

  @Consumes("application/json")
  @Produces("application/json")
  @POST @Path("/ipa/sessions/json")
  @CustomJsonRpcAnno("user_delete")
  UserDeleteResponse userDelete(UserDeleteParams params);
}
```

See how they have the exact same path?  The only way that JSON rpc (http://json-rpc.org/) knows what action to take is based on a special parameter "method".  I use the `WebResourceInvocationCallback` to look for a `CustomJsonRpcAnno`, grab the value, and add it as a request context property which can be read in a message body writer.

Again, if there are other ideas, happy to implement them.  All that is needed is someway to get a handle on annotations which were on the client interface in a jax-rs interceptor

PR #257, opened 2017-01-24
Jetty HTTP/2 client connector

nhenneaux:
Provide a HTTP/2 client connector for Jersey


pavelbucek:
Hi Nicolas,

thanks for your contribution!

To be able to process your pull request, you have to sign OCA: http://www.oracle.com/technetwork/community/oca-486395.html.

Thanks and regards,
Pavel

nhenneaux:
Hi,

I have just sent the OCA completed.

Nicolas

pavelbucek:
How is this different from exiting jetty-connector? Couldn't we have both (http and http2) supported by the single one?

nhenneaux:
For the moment there is no connector supporting HTTP/1.1 and HTTP/2 at the same time. I have made an [improvement request on the Jetty GitHub](https://github.com/eclipse/jetty.project/issues/1350) but for the moment, we have to choose one.

nhenneaux:
For the moment there is no connector supporting HTTP/1.1 and HTTP/2 at the same time. I have made an [improvement request on the Jetty GitHub](https://github.com/eclipse/jetty.project/issues/1350) but for the moment, we have to choose one.

pavelbucek:
That would mean having two client instances (both with different connector) when we want to talk with http and http2. Also, how can user know in advance which server supports which http version?

If you don't mind, I'd better wait for Jetty to provide single client, which can handle multiple http versions..

nhenneaux:
You usually know what the server supports. If a server starts to support
HTTP/2 it will hardly not go back to HTTP/1.

I am currently using the HTTP/2 transport anyway so I don't mind to wait.
The idea of this PR was only to share the implementation with the community.

pavelbucek:
I don't agree - for example when accessing load-balanced service, it can happen that you'll get redirected to a node which supports only http 1.x or to a node which supports http/2. This was very common in my part of the world when accessing google services.

nhenneaux:
In this setup, you cannot use Jersey anyway for the moment. This is only a
first step to get something working for HTTP/2 servers.

pavelbucek:
Ok.

In that case, let me close this pull request for now. The main issue here is that this is http/2 only client with limited usability, which forces users to do much more than we want them to do - know about the transport used when accessing some endpoint.

Also, the change as it is now introduces lots of duplicate code - that's something we want to avoid if possible.

Feel free to create your github repo with this connector or something like that and if you want to promote it, feel free to share it on users@jersey.java.net mailing list.

Thanks for your contribution! I hope this will make it in, when jetty is updated to support both (or when connector will be able to do that without having user to decide).

PR #235, opened 2016-08-24
JERSEY-1653: Add support for @BeanParam in jersey-proxy-client

lucastsa:
This pull requests adds support to implement proxy clients on resources that use JAX-RS `@BeanParam` annotation.

Main change is introduced in the new method `extractParamsFromBeanParamClass()` that handle bean class info. Most of the rest is moving logic from the main invoke procedure to separate method for reuse.

I'm creating this on 2.x branch so it is released as part of future Jersey 2 releases, but I think the extension is exactly the same on master now.

PR #259, opened 2017-02-04
Add BeanParam support to WebResourceFactory

daicoden:
This allows you to use BeanParam to consolidate argument lists for requests with many QueryParams (or other types of parameters).


daicoden:
Oh man - I just realized #235 does the same thing... Any reason this hasn't got any movement?

pavelbucek:
Nether of those requests do have an OCA signed, thus we cannot move it forward:

Best regards,
Pavel

daicoden:
Signed OCA, closing in favor of: https://github.com/jersey/jersey/pull/235 since it looks like @lucastsa signed it too.

PR #260, opened 2017-02-07
Logging : Max entity size limit check

thegamblerrises:
If a LoggingFeature is registered and provided value for maxEntitySize is set to Integer.MAX_VALUE. Then at runtime, jersey logging throws `NegativeArrayException` which is because of this [line](https://github.com/jersey/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/logging/LoggingInterceptor.java#L210)

Since, jersey logging treats negative values as 0 and it doesn't log more than `DEFAULT_MAX_ENTITY_SIZE`. So, a check for max value would be great. I read that logging shouldn't cause problem to main call flow. So, here is a suggestion for suppressing the provided `maxEntitySize`.


pavelbucek:
Thanks for your contribution!

Best regards,
Pavel

PR #261, opened 2017-02-09
Fixes error where Netty was not binding properly to a host.

ccleve:
The NettyHttpContainerProvider currently ignores the "host" part of the baseUri. This is ok when you want Netty to bind to the main network interface on a server, but it's a problem when there are multiple interfaces, or you want to do loopback testing. For example, you might want to have one instance of Netty bind to 127.0.0.1 and a second one bind to 127.0.0.2 so you can test communications between the two.


pavelbucek:
Thanks for your contribution!

Best regards,
Pavel

ccleve:
I just send the OCA.

ccleve:
Any thoughts on this? I'd like to get rid of my custom version of Jersey.

PR #262, opened 2017-02-11
Supporting hashCode and equals calls on jersey client proxy objects.

wadechandler:
Spring Boot 1.4+ and some other things have issues because hashCode and equals will throw unsupported exceptions. I have signed the OCA a long time back; you can see my name in the list. Please review and merge.

This will also need to be back ported to 2.x. Do I need to do that and submit a PR? If so, what is the normal flow for that in the Jersey project.


wadechandler:
If I need to port to 2.x, what is the best branch? Would it be `2.x` or all or a subset of `2.*.x` branches?

pavelbucek:
merged to 2.x and master.

wadechandler:
Thanks much @pavelbucek

PR #267, opened 2017-03-17
Jersey-3214

MarkusKull:
https://java.net/jira/browse/JERSEY-3214  

Bugfix to prevent exception when closing unconsumed Response. Affected only httpclient 4.5.1+ . Jersey currently depends on unaffected httpclient 4.5. Also reintroduces the older behaviour to consume the response when closing content-inputstream. No change when closing response: connection is discarded without consuming entity.

OCA is signed for 1&1 (literally last entry in list)

PR #268, opened 2017-03-20
Apache connector: Removed unnecessary InputStream.close() call that causes exception on httpclient 4.5.2

agherardi:
I believe this should fix https://java.net/jira/browse/JERSEY-3214


connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -617,18 +617,6 @@ public boolean isStreaming() {
...
-        HttpClientResponseInputStream(final InputStream inputStream) throws IOException {
-            super(inputStream);
-        }
-
-        @Override
-        public void close() throws IOException {
-            super.close();
-        }
-    }
-
```
agherardi: The only FilterInputStream method that this class overrides is close(). However the override close() method calls the superclass' close() method, so it's unnecessary.

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -648,7 +636,6 @@ private static InputStream getInputStream(final CloseableHttpResponse response)
             @Override
             public void close() throws IOException {
                 response.close();
-                super.close();
             }
```
agherardi: See httpclient documentation. To properly dispose of resources, one must either close the CloseableHttpResponse or the InputStream. Closing both is unnecessary and causes an exception in httpclient 4.5.2.

MarkusKull:
Oh. There is already #267 which also tries to address this problem.

Difference between the two Pull Requests: #268 always aborts the connection (when not already consumed). #267 aborts the connection when closing the response, but consumes when closing the inputstream (identical behaviour to apache httpclient).

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
aaronjwhiteside: Calling `response.close()` causes no connections to be released back to the pool.

https://java.net/jira/browse/JERSEY-3260

agherardi: If the application consumes the entity, by the time response.close() runs the connection has already released to the pool, so response.close() does nothing.

If the application does NOT consume the entity, there is no other choice but closing the connection - reusing it is not possible in that case.

agherardi:
If the application closes the inputstream, most likely the application is not interested in the response content. In that case, aborting the connection is the best course of action - consuming the response blocks the application until it receives the entire content.

To summarize: 
* If the application is interested in the response content, the application can consume the entity then close the response. Barring other issues, the connection is reusable and released to the pool.
* If the application is not interested in the response context, the application can immediately close the response. In this case, the connection is aborted and cannot be reused. That's better than blocking the application in an attempt to reuse the connection.

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
aaronjwhiteside: @agherardi I agree with you. However `response.close()` will always cause the connection to be closed regardless of whether or not the entity was read in its entirety. Follow the chain of calls that response.close() makes and you will see it eventually calls `ConnectionHolder.releaseConnection(false)` - the `false` bit means the connection should not be reused (returned to the pool).

```
HttpClientResponseInputStream.close() -> CloseableHttpResponse.close() -> HttpResponseProxy.close() -> ConnectionHolder.close() -> ConnectionHolder.releaseConnection(false)
```

There needs to be some conditional check to see if the entity has not been consumed yet and then call `response.close()` otherwise call `super.close()`. 

Another alternative could be to have an option to consume/read any remaining bytes of the entity and discard them, I'm not sure this should be the default behavior but it seems very useful.

agherardi: >  ConnectionHolder.releaseConnection(false)  - the  false  bit means the connection should not be reused (returned to the pool).

True. But if you check ConnectionHolder.releaseConnection(final boolean reusable), notice that it enters the block that aborts the connection only if:

`if (this.released.compareAndSet(false, true)) {`

When I test this code under debugger, if the application has already consumed the entity, this.released is already set to TRUE so the block of code that aborts the connection doesn't run.

agherardi: Another way to verify that my patch does NOT break connection reuse is running the code below - with my patch - while wiresharking traffic. Assuming the server honors Connection: Keep-alive, you'll see that both GET requests use the same TCP/HTTP connection.

    public static void main(String[] args) throws Exception
    {
      Client client = getClient();
       doJersey(client);
       doJersey(client);
     }
    
     public static void doJersey(Client client) throws Exception
     {
       String url = "http://...";
    
       WebTarget resource = client.target(url);
       Response response = resource
        .request()
        .get();
    
       response.readEntity(String.class);
       response.close();
     }

aaronjwhiteside: Interesting, my use case uses `WebResourceFactory` for proxy generation, where the interface method returns a Model (so the assumption is that if the model is returned, the entity must be consumed), maybe the bug has something to do with the code in the proxy.

aaronjwhiteside: So I've narrowed this down to be an issue with the proxies, if I don't use a proxy like in your example everything is a-ok.

Thanks again for your time, I think I'll go bother someone else now :)

agherardi:
OCA is here http://www.oracle.com/technetwork/community/oca-486395.html

agherardi:
Re-opened against master as https://github.com/jersey/jersey/pull/3565 .

PR #271, opened 2017-04-21
2.6.x

finrod2002:
Backport of  JERSEY-2125 fix into the 2.6.x branch, so that the fix is available for people that still have to comply to Java 1.6 source level.


finrod2002:
I needed the fix JERSEY-2125 into a Jersey version that can run on Websphere 7.0, which only supports Java SE 6. Therefore I needed to backport the fix into the 2.6 branch.

finrod2002:
Sorry, for the change in the pom.xml. I see it only now, this needs to be reverted.

PR #3545, opened 2017-04-26
JERSEY-3258: Fixed error handling for async client

alex-shpak:
Hi!

PR addressing [JERSEY-3258](https://java.net/jira/browse/JERSEY-3258) bug.
Exceptions thrown while processing response (for example JSON parsing) in async client was not handled properly and was leading client to wait forever.

OCA is signed and sent.


pavelbucek:
Hi Alex,

thanks for your contribution.

In order to keep our code coverage at reasonable level, we do require a test for each bugfix - please updage your PR to include at least some sanity test.

Best regards,
Pavel

alex-shpak:
Hi! Sure.

I added test. Without changed line test will hang forever.

alex-shpak:
Also added timeout just in case, so test would not hang in CI

alex-shpak:

PR #3552, opened 2017-05-12
Avoid BufferCapacity exceeded error on large http response

daijithegeek:
This fix is to avoid exception when the response content overflows the responseBufferSize.

I did not expose the responseBufferSize setting through the properties mechanism as it will not fix the problem, but only report it. This is an addition that can be done though.

Stop sending http requests by using simple send() function in order to avoid BufferCapacity exceeded error. Instead, using send with a listener that allows response exceeding the buffer capacity to be processed, if not exceeding timeout.

### DIFF
--- a/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
+++ b/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
@@ -67,6 +67,7 @@ import javax.ws.rs.core.MultivaluedMap;
 
 import javax.net.ssl.SSLContext;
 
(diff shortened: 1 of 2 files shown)


pavelbucek:
can you please add test for your change?

Also, we cannot evaluate and accept your proposal without you signing Oracle Contributor Agreement: http://www.oracle.com/technetwork/community/oca-486395.html. Please let us know whether you plan to do that.

daijithegeek:
OCA is signed and sent.

I added a test that fails in default, but works with the PR fix.

delabassee:
@daijithegeek I don't find your OCA?! Can you ping me directly david.delabassee(at)oracle.com to fix this?

PR #3563, opened 2017-05-24
Allow the same Jersey client to be used for both authenticated and non-authenticated URLs

agherardi:
When registering the basic HttpAuthenticationFeature, the BasicAuthenticator shouldn't throw an exception when credentials are missing, as the request could be for a non-authenticated resource.

This allows the same Jersey client to be used for both authenticated and non-authenticated URLs.


agherardi:

agherardi:
Any updates? Thanks

agherardi:
Any updates?

Issue #3506, opened 2017-02-20
InetAddresses.isUriInetAddress(String) from the emdeded guava lib could be optimized


glassfishrobot:
JerseyUriBuilder.host(String) employs InetAddresses.isUriInetAddress(String) that is embedded in jersey (repackaged or moved to uri.internal). The problem with InetAddresses.isUriInetAddress is that:
1- The flow is controlled by throwing an IllegalArgumentException which may cause a performance hit if the stack is deep (fillinstacktrace).
2- The message is passed to the exception using String.format which is also an uncalled for performance hit.
The remedy is quite simple, either change the guava dependency from 18 to 20 (it was fixed in that version) or just change the isUriInetAddress and forUriString to control the flow in a better way rather than using exceptions
#### Affected Versions
[2.22.2, 2.25.1]

dreis2211:
+1

PR #3577, opened 2017-06-08
3506: Avoid exception control flow in isUriInetAddress

schlosna:
Port of upstream [Guava commit e6003692500352e1ba59eee5c93bc21acb3f46c0](https://github.com/google/guava/commit/e6003692500352e1ba59eee5c93bc21acb3f46c0) which avoids using exceptions as control flow in Jersey's internal version of InetAddresses.isUriInetAddress().

This avoids the expense of exception creation that can impede request throughput, especially under high load, and addresses #3506 and previously JERSEY-3234.


schlosna:
I have previously signed [OCA](http://www.oracle.com/technetwork/community/oca-486395.html).

pavelbucek:
Thanks!

Issue #3593, opened 2017-06-24
WebResourceFactory issue when @Context after entity in signature


boubakov:
When a resource is defined such as an entity is defined before an `@Context` parameter (see ex below), `WebResourceFactory` fails to set the entity

`@Path("postEntityAndContext")
    @POST
    @Consumes({MediaType.APPLICATION_XML})
    @Produces({MediaType.APPLICATION_XML})
    MyBean postEntityAndContext(@Valid MyBean entity, @Context SecurityContext securityContext);`

PR #3594, opened 2017-06-24
#3593: WebResourceFactory issue when @Context after entity in signature

boubakov:
Fixed JERSEY#3593. `@Context` annotated parameters are now ignored in `WebResourceFactory` such that entity is not lost anymore.
Test added.
OCA already sent.


boubakov:
Email containing OCA sent on 24.06. No feedback at the moment.

delabassee:
@pavelbucek @boubakov OCA has been reviewed and approved!

PR #3596, opened 2017-06-27
Fix sample README default URLs

freddyboucher:


pavelbucek:
Thanks for the PR.

I'm closing it since you don't have an OCA and also that the content fits into another commit which is under way.

PR #3600, opened 2017-06-30
Adding support for opening EventSource with a pre-known eventId

JoergAdler:
Hi,

I tried to use a jersey client with an SSE stream in following scenario: 

The client subscribes to the stream and persist the current eventId to a database.
After a restart, the client reads the persisted last id from the database and wants to reconnect to the stream with the pointer to the current event. This feature is already supported if the client looses his connection but unfortunately not for restarting the whole client. 

I added another open() method to the EventSource class, where you can pass the eventId to start from. 

It would be nice, if you could consider to merge this.

Thanks a lot.

### DIFF
--- a/media/sse/src/main/java/org/glassfish/jersey/media/sse/EventSource.java
+++ b/media/sse/src/main/java/org/glassfish/jersey/media/sse/EventSource.java
@@ -283,6 +283,20 @@ public class EventSource implements EventListener {
      * @throws IllegalStateException in case the event source has already been opened earlier.
      */
     public void open() {
+        open(null);
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


pavelbucek:
Hi Joerg,

thanks for the PR. We can't do anything with it unless you sign Oracle Contributor Agreement (OCA): http://www.oracle.com/technetwork/community/oca-486395.html

please let us know whether you plan to do that.

Regards,
Pavel

JoergAdler:
Hi Pavel,

I just sent the OCA to you :-)

Greetings
Jörg

JoergAdler:
Hi @pavelbucek ,

is there something new in this case?

Greetings
Jörg

PR #3612, opened 2017-07-06
Fixes #3609

### LINKED ISSUE #3609: JettyConnector sends 2 User-Agent headers
Hi all,

in [JettyConnector.java](https://github.com/jersey/jersey/blob/311cc47e7d83f86b925cf60700075753dbdb5387/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java#L245) line 245, the jetty request is created with 2 headers by default : 

screen shot 2017-07-06 at 19 05 40

The jersey ClientRequest also contains a User-Agent : 

screen shot 2017-07-06 at 19 07 21

in [JettyConnector.java](https://github.com/jersey/jersey/blob/311cc47e7d83f86b925cf60700075753dbdb5387/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java#L317) line 317, the headers are merged, resulting in :

screen shot 2017-07-06 at 19 09 42

Microsoft IIS 10 does not accept this and returns a 400.

hypnoce:
Fixes #3609  : Combine replicated header in JettyConnector.

### DIFF
--- a/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
+++ b/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
@@ -63,6 +63,7 @@ import java.util.logging.Logger;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.core.Configuration;
+import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.MultivaluedMap;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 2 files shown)


hypnoce:
OCA signed

AdamLindenthal:
Hi Francois, 

thanks for sumbitting the pull request. However, we solved the issue via internal [commit](https://github.com/jersey/jersey/commit/3ab55c249f9b86169e3ae12bd6494d673a9fef5e). Your patch would create a header containing multiple values. The `User-agent` header added by Jersey already contains the information about the underlying Jetty version. So, in the end, we went with a dead simple solution, as can be seen in the commit.

Regards,
Adam

hypnoce:
Thanks Adam,

I agree it will be a multiple value header. Strictly speaking, this is legal as per the http specification. This PR ment to provide a more general solution that does not rely on the fact the jetty has default headers.
Thanks for fixing it.

Regards

Francois

PR #3616, opened 2017-07-12
Fixes #3613.

### LINKED ISSUE #3613: OAuth1ServerFilter does not work with cdi2-se
Using the new cdi2 (version 2.26-b7) instead of hk2 for injection, OAuth1ServerFilter does not work anymore. The uriInfo is injected with @Inject annotation which does not follow the requirements from https://jersey.github.io/documentation/latest/jaxrs-resources.html#d0e2752. 
Here a test reproducing the issue : https://github.com/hypnoce/jersey/commit/eec78704e936ba8128bc09b4bbad3d3353274c4d

hypnoce:
I multiple SupplierBeanBridge (true for all Supplier*Bridge) are registered in a request scope and injected as proxy, they occupy the same key in the BeanStore. Therefore subsequent bridges will be backed with the intense of the first and lead to illegal calls.
The fix does not cover the cases for other bridges.


hypnoce:
OCA signed

PR #3669, opened 2017-09-14
ParamConverterFactory should use LinkedHashSet to preserve ordering

chkal:
The class `ParamConverterFactory` manages a list of `ParamConverterProvider` implementations. But unfortunately the order in which the providers are processed seems to change between deployments. The reason for this is that the factory uses a `HashSet` internally. The provider sets passed into the constructor are `LinkedHashSet` instances, but as they are copied to the `HashSet`, the order may change.


mkarg:
@pavelbucek As ordering of components is essential for applications to work correctly, I would kindly like to ask to include this PR in the next bugfix release of Jersey. :-)

PR #3683, opened 2017-09-25
Register providers added in `Feature`s

### LINKED ISSUE #3675: HK2 binder registered in Feature is ignored
### Description

Given [the major breaking change around Jersey DI](https://jersey.github.io/documentation/latest/migration.html#mig-2.26), I have spent some time today in the attempt to migrate a simple web app based on Jersey 2.25.1 to Jersey 2.26.

The web app contains a `ResourceConfig` which registers:
1. a `org.glassfish.hk2.utilities.binding.AbstractBinder`;
2. a `javax.ws.rs.core.Feature` which, in turn, registers another `org.glassfish.hk2.utilities.binding.AbstractBinder`.

Since I'm heavily relying on HK2-specific annotations, I tried to add `jersey-hk2` as a dependency of the web app but the following came out:
* the binder at _1_ is configured (i.e. its `configure` method is invoked during startup);
* the binder at _2_ is **not** configured (i.e. its `configure` method is **not** invoked during startup).

Also, if I switch `org.glassfish.hk2.utilities.binding.AbstractBinder` with `org.glassfish.jersey.internal.inject` **only in 2**, that binder is properly configured instead.

I guess (and I hope) that depending on `jersey-hk2` should allow me to consistently register HK2 binders either directly on a `ResourceConfig` and indirectly through `Feature`s.

fabriziocucci:
Because providers registered in `Feature`s were not considered, registering a HK2 binder in a `Feature` didn't work.

This fixes #3675.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java
+++ b/core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java
@@ -52,12 +52,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
+import javax.annotation.Priority;
 import javax.ws.rs.ConstrainedTo;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.RuntimeType;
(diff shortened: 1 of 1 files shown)


fabriziocucci:
I'm not sure if this is the best way to fix #3675 and, actually, I'm not even sure whether the Jersey team consider #3675 a bug since the issue has not been labeled yet.

But if it is, a proper test needs to be added. I honestly wasn't sure where to place it. Maybe in `core-common`? Or perhaps in `tests/integration/jersey-3683`? Or somewhere else?

Issue #3721, opened 2017-11-21
Undertow container implementation


jcomo:
Not sure what the protocol is for accepting container implementations, but I have written one for [Undertow](http://undertow.io/) that others might find useful. If this is something you'd be interested in taking a look at, I can open a PR with the patch

PR #3722, opened 2017-11-21
Undertow container implementation

jcomo:
This is the code implementation for #3721. I'm open to discussion about this. It stemmed from a need on our end, but I decided to push it upstream. I based this implementation off of the other containers, but it would be great to have someone look over it with more context (as this is also my first time contributing to the library) to see if I've missed any key points. I _think_ I followed style for comments, localization, etc. but there was no contributing guide.

Note that I branched this off of 2.26 since I was having issues getting master building on my machines.

### DIFF
--- a/containers/undertow-http/pom.xml
+++ b/containers/undertow-http/pom.xml
@@ -45,59 +45,55 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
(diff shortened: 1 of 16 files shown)


jcomo:
@pavelbucek is this something you would be interested in taking a look at?

Issue #3293, opened 2015-12-11
HttpUrlConnector.secureConnection not always setting custom SSL socket factory


glassfishrobot:
To be more precise, due to changes introduced by #3171 ([https://github.com/jersey/jersey/commit/57f5dafed38d27fcb70e7f4ee5051554118d305a#diff-2af8c9e0f0f9963b8b1ce356cf17ecb7](https://github.com/jersey/jersey/commit/57f5dafed38d27fcb70e7f4ee5051554118d305a#diff-2af8c9e0f0f9963b8b1ce356cf17ecb7)), a custom SSL socket factory is not set on an HttpsURLConnection 'suc' in HttpUrlConnector unless HttpsURLConnection.getDefaultSSLSocketFactory returns the exact same object as suc.getSSLSocketFactory(). This is not a robust way of determining if the custom socket factory has already been set because HttpsURLConnection.getDefaultSSLSocketFactory() can be called simultaneously by multiple threads, returning a different object for each invocation, thus resulting in up to n-1 threads getting a mismatch in HttpUrlConnector.secureConnection, even though suc.getSSLSocketFactory() is actually returning a default SSL socket factory (just not necessarily the exact instance we expected, which is ultimately irrelevant).

This is fairly easy to reproduce, given an HTTPS server with a self-signed certificate. The following code is not a self-contained unit test, but will demonstrate the issue nevertheless, by spitting out at least one SSLHandshakeException (caused by the custom SSLContext not being used). If numThreads is set to 1, then it will never fail due to a certificate validation error. The code will also run without error on Jersey 2.21 or earlier.

```
import javax.net.ssl.*;
import javax.ws.rs.client.Client;
import javax.ws.rs.client.ClientBuilder;
import java.security.SecureRandom;
import java.security.cert.CertificateException;
import java.security.cert.X509Certificate;

public class HttpsTest {

    private static SSLContext sslContext = null;

    private static synchronized SSLContext getSslContext() {

        if (sslContext == null) {
            try {
sslContext = SSLContext.getInstance("TLS");
sslContext.init(null, new TrustManager[]{new X509TrustManager() {
    public void checkClientTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
    }

    public void checkServerTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
    }

    public X509Certificate[] getAcceptedIssuers() {
        return new X509Certificate[0];
    }

}}, new SecureRandom());

            } catch (Exception e) {
[...]

PR #3738, opened 2017-12-19
#3293 Pre-load default ssl socket factory in HttpUrlConnector

kevinconaway:
@mpotociar @petrbouda this PR is a potential fix for #3293

From my understanding of the current code, the logic for using a custom SSL socket factory is:

1. If the `HttpsUrlConnection` returned by the `ConnectionFactory` already has a custom `SSLSocketFactory` set on it, do nothing
1. Otherwise, set the socket factory on the connection to be the one provided from the `SSLContext` in the client config.

Currently, the [first check](https://github.com/jersey/jersey/blob/2.26/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java#L325) is done via:

```java
if (HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory()) {
    // indicates that the custom socket factory was not set
    suc.setSSLSocketFactory(sslSocketFactory.get());
}
```

The issue with this code is that `HttpsURLConnection.getDefaultSSLSocketFactory() ` is not thread safe.  The implementation looks like:

```java
    public static SSLSocketFactory getDefaultSSLSocketFactory() {
        if (defaultSSLSocketFactory == null) {
            defaultSSLSocketFactory = (SSLSocketFactory)SSLSocketFactory.getDefault();
        }

        return defaultSSLSocketFactory;
    }
```

If there are concurrent requests, one of them may end up creating a separate `SSLSocketFactory` that is not the default.  The comparison check above fails and the request ends up using the default SSL context instead of the provided one.

The fix in this PR is to call `HttpsURLConnection.getDefaultSSLSocketFactory()` in a static initializer block for the class so that the default SSLContext is present when the comparison is made in `HttpUrlConnector#secureConnector`

I'm open to feedback on alternate approaches to resolving this.

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
@@ -40,6 +40,28 @@
 
 package org.glassfish.jersey.client.internal;
 
+import org.glassfish.jersey.client.ClientProperties;
+import org.glassfish.jersey.client.ClientRequest;
+import org.glassfish.jersey.client.ClientResponse;
+import org.glassfish.jersey.client.HttpUrlConnectorProvider;
+import org.glassfish.jersey.client.JerseyClient;
+import org.glassfish.jersey.client.RequestEntityProcessing;
+import org.glassfish.jersey.client.spi.AsyncConnectorCallback;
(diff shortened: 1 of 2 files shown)


kevinconaway:
@pavelbucek are you the correct person to review this?

PR #3546, opened 2017-05-01
Patch for client connection leak when using digest authentication

agherardi:

PR #3751, opened 2018-01-08
Patch for client connection leak when using digest authentication

agherardi:


agherardi:

agherardi:
Replaces https://github.com/jersey/jersey/pull/3546

PR #3565, opened 2017-05-27
Apache connector: Removed unnecessary InputStream.close() call that causes exception on httpclient 4.5.2

agherardi:
See https://github.com/jersey/jersey/pull/268

PR #3752, opened 2018-01-08
Properly close the Apache response so that connections can be reused

agherardi:


agherardi:
Obsoletes https://github.com/jersey/jersey/pull/3565

agherardi:
See http://hc.apache.org/httpcomponents-client-4.5.x/tutorial/html/fundamentals.html#d5e145

> In order to ensure proper release of system resources one must close either the content stream associated with the entity or the response itself
> The difference between closing the content stream and closing the response is that the former will attempt to keep the underlying connection alive by consuming the entity content while the latter immediately shuts down and discards the connection.

PR #3777, opened 2018-02-24
Allow external modifier while building HttpClient

benmingli:
Hi,

In our use case, we need to modify the HttpClientBuilder (regarding SSL Context, and misc) and also build the client using our own way (instead of using HttpClientBuilder.build() )..

Please review my code and let me know if this works.

Thanks a lot,
Benjamin.


benmingli:
Please see the feature request 
https://github.com/jersey/jersey/issues/3778

Issue #3338, opened 2016-03-01
JettyHttpContainer URL handling broken


glassfishrobot:
I'm creating a JAX-RS server using the jetty http container. The resource
is annotated with @Path("/services") and the whole JAX-RS handler is put under context "/my/path" (see code snippet below for details)

So an HTTP GET on <server:port>/my/path/service should end up in my rest resource. However, it does not on 2.22.2\.

After stepping through the code, it seems that the JettyHttpContainer.handle constructs a wrong requestUri:"http://localhost:52021/my/path/my/path/services"
(notice the duplication of /my/path)

It is related to fixed done in scope of [https://github.com/jersey/jersey/commit/239d02fcc0792155eb5aaf51e5c2b68e61203c19#diff-1997febdf388bca1dd941929123602df](https://github.com/jersey/jersey/commit/239d02fcc0792155eb5aaf51e5c2b68e61203c19#diff-1997febdf388bca1dd941929123602df)

I have reverted back to 2.21.1, which works fine.

```
ContextHandler context = new ContextHandler();
        final Handler endpoint = RuntimeDelegate.getInstance().createEndpoint(application, Handler.class);

        context.setHandler(endpoint);
        context.setContextPath("/my/path");
```
#### Affected Versions
[2.22.2]

brettkail-wk:
As a workaround, if you're willing to use jetty-servlet, you can use:
```
ServletContextHandler servletContextHandler = new ServletContextHandler(null, "/path");
servletContextHandler.addServlet(new ServletHolder(new ServletContainer(config)), "/*");
server.setHandler(servletContextHandler);
```

PR #3786, opened 2018-03-20
Fix Jetty context path

brettkail-wk:
`JettyHttpContainer` has logic to support non-default context paths, but it miscalculates the request URI: the `getRequestURI` contains the context path, and it is appended `getServerAddress(baseUri)`, which also (correctly) contains the context path.  Use `getPathInfo` instead, which excludes the context path.

(I am listed for Jersey in the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html#w) under "Workiva Inc.")


konrader:
Sounds like this would fix https://github.com/jersey/jersey/issues/3338

brettkail-wk:
@konrader Yes, that appears to be the same issue.

